PR #11, opened 2021-04-23
Implement Sign in view

COSMO-VMI:
- Update translate file
- Move translation from App to SignIn
- Remove obsolete dev and static auth
- Implement sign in view with new look and feel, and use SignInButton from ui repo

### DIFF
--- a/src/views/SignIn/SignIn.js
+++ b/src/views/SignIn/SignIn.js
@@ -3,16 +3,22 @@
 
 import React, { useState, useEffect } from 'react'
 import { withRouter } from 'react-router-dom'
-import { useTranslation } from 'react-i18next'
+import { Trans, useTranslation } from 'react-i18next'
 import { Auth, AuthDev } from '@cosmotech/core'
-import { AuthMSAL, AuthStaticWebApp } from '@cosmotech/azure'
+import { AuthMSAL } from '@cosmotech/azure'
 import validate from 'validate.js'
 import { makeStyles } from '@material-ui/styles'
 import {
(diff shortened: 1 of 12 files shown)


src/views/SignIn/SignIn.js
```diff
@@ -81,55 +87,65 @@ const useStyles = makeStyles(theme => ({
...
 }))
 
-const SignIn = (props) => {
-  const { t } = useTranslation()
+const SignIn = () => {
+  const { t, i18n } = useTranslation()
 
   const classes = useStyles()
 
+  const changeLanguage = (language) => {
```
jreynard-code, line 135: It can be interesting to extract this method inside a utility file. I think about the UserInfo component which certainly could reuse these lines.

src/views/SignIn/SignIn.js
```diff
@@ -189,43 +199,64 @@ const SignIn = (props) => {
...
+                      onChange={(event) => changeLanguage(event.target.value)}
+                    >
+                      <MenuItem value={'en'}>English</MenuItem>
+                      <MenuItem value={'fr'}>Français</MenuItem>
+                    </Select>
+                  </FormControl>
+                </Grid>
+                <Grid item>
+                  <Typography color="textSecondary" variant="caption">
+                    © 2021 Cosmo Tech
```
jreynard-code, line 255: Extract the label here (maybe partners will want to ovewrite this.
 Also, use <Trans /> from react-i18next in order to format the date from JS call instead of hard-coded value

PR #14, opened 2021-05-03
Add map of module references for dependencies of linked modules

csm-thu:
Jest tests fail locally when `yarn link` points to a lib with
peer dependencies. This map configuration adds redirection to
import these dependencies from <rootDir>/node_modules instead
of the linked modules folders.

https://github.com/facebook/jest/issues/2447#issuecomment-297766714


jreynard-code (approved):
LGTM ;)

PR #16, opened 2021-05-11
NBO : Scenario creation and selection

jreynard-code:
Add Scenario dropdown in Scenario view
Add Create scenario button and create scenario dialog in Scenario view
Remove /api folder and add /src/services with [cosmotech-api-js-client ](https://github.com/Cosmo-Tech/cosmotech-api-javascript-client) use
Add cosmotech-api-javascript-client configuration in ./configs/Api.config.js
Add Worskpace and organisation notions
Add solution, workspace, runTemplates into redux
Adapt saga to use service layer instead of axios lib
Add Worskpace, dataset and solution loadinglines in Loading  Screen 

@csm-thu @Nikeleos @FMO-Cosmo for information


COSMO-VMI (commented):
Apart from the aforementioned comments, LGTM

PR #29, opened 2021-05-20
VMI/number validator prod 7582

COSMO-VMI:


csm-thu (commented):
Hello ! The dependency _react-number-format_ doesn't seem to be used. Can we remove it ?

Nikeleos (approved):

COSMO-VMI:
> Hello ! The dependency _react-number-format_ doesn't seem to be used. Can we remove it ?

Oh, I added the dependency here for the number input component, but I should add it in the ui repo, which contains the component, right?

csm-thu:
> Oh, I added the dependency here for the number input component, but I should add it in the ui repo, which contains the component, right?

That's correct, this dependency is only used in the component located in the UI repository, and since it is a direct dependency (i.e. not a peer dependency) it does not need to be added in the webapp repository :)

csm-thu (approved):
LGTM :+1:

PR #33, opened 2021-05-24
Thu/update scenario parameters prod 7559

csm-thu:
Finish connecting button "Update & Launch" to the API REST, retrieve and store the simulation run id.

### DIFF
--- a/src/services/scenario/ScenarioService.js
+++ b/src/services/scenario/ScenarioService.js
@@ -3,8 +3,8 @@
 
 import { CosmotechApiService } from '../../configs/Api.config';
 import {
-  formatScenarioParametersForApi,
-  formatScenarioParametersFromApi
+  formatParametersForApi,
+  formatParametersFromApi
 } from '../../utils/ApiUtils';
 
 const ScenarioApi = new CosmotechApiService.ScenarioApi();
@@ -42,10 +42,10 @@ function createScenario (organizationId, workspaceId, scenario) {
 }
 
 function updateScenarioParameters (organizationId, workspaceId, scenarioId, scenarioParameters) {
-  const formattedParameters = formatScenarioParametersForApi(scenarioParameters);
+  const formattedParameters = formatParametersForApi(scenarioParameters);
   return new Promise((resolve) => {
     ScenarioApi.addOrReplaceScenarioParameterValues(organizationId, workspaceId, scenarioId, formattedParameters, (error, data, response) => {
-      data = formatScenarioParametersFromApi(data);
+      data = formatParametersFromApi(data);
       resolve({ error, data, response });
     });
   });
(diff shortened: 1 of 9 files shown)


src/services/scenario/ScenarioService.js
```diff
@@ -37,9 +39,40 @@ function createScenario (organizationId, workspaceId, scenario) {
   });
 }
 
-function updateAndLaunchScenario (organizationId, workspaceId, scenariodId, scenarioParameters) {
+function formatScenarioParametersForApi (scenarioParameters) {
```
jreynard-code, line 42: Should be placed in utility file.
We will certainly reuse the logic here somewhere else :thinking:

src/services/scenario/ScenarioService.js
```diff
@@ -37,9 +39,40 @@ function createScenario (organizationId, workspaceId, scenario) {
...
+    const newParam = clone(param);
+    // Cast boolean values into string values
+    if (newParam.varType === 'bool') {
+      newParam.value = newParam.value.toString();
+    }
+    return newParam;
+  });
+}
+
+function formatScenarioParametersFromApi (scenarioParameters) {
```
jreynard-code, line 55: Should be placed in utility file.
We will certainly reuse the logic here somewhere else :thinking:

src/state/sagas/scenario/UpdateAndLaunchScenario/UpdateAndLaunchScenario.js
```diff
@@ -1,34 +1,63 @@
...
+    } else {
+      // Get csmSimulationRun id
+      const csmSimulationRun = runData.csmSimulationRun;
+      if (csmSimulationRun === undefined) {
+        console.error('csmSimulationRun is undefined');
+      } else {
+        yield put({
+          type: SCENARIO_ACTIONS_KEY.SET_CURRENT_SCENARIO,
+          data: {
+            status: STATUSES.SUCCESS,
```
jreynard-code, line 46: Should be RUNNING instead of SUCCESS.
At the moment, the simulation is running but is not necessarily finished.
The RUNNING status will allow us to display "Scenario running message" instead of ScenarioResultIframe, disable Scenario run button,... during a scenario run.
BUT, because we cannot know, for the moment, when a scenario launched, is finished, we need to trigger each X seconds (configurable in App.config.js) the scenario run status

jreynard-code (changes requested):
Minor changes except for UpdateAndLaunchScenario.
Maybe we should discuss with @FMO-Cosmo to be sure if it's needed in V1.0.0

src/state/sagas/scenario/UpdateAndLaunchScenario/UpdateAndLaunchScenario.js
csm-thu: As discussed previously, this will be handled in PROD-7326 (branch THU/run_in_progress_PROD-7326)

PR #55, opened 2021-05-31
THU/auth_access_token_PROD_7615

csm-thu:
Set the access token, if available, in the cosmotech API service.
If the access token is not available after login (e.g. when using the dev mode), the service token is set to "none".

Access token is retrieved from local storage.


jreynard-code (approved):
LGTM 😉

PR #68, opened 2021-06-08
THU/scenario_management_scripts

csm-thu:
Add a python script to make scenarios removal easier.
Note: this script does not handle the change of scenarios' parent id yet.


csm-thu:
These scripts are DevOps tools and will be placed in a different repo to keep this one focused on the webapp.
Closing the PR.

PR #72, opened 2021-06-16
feat: Implement util functions to sort scenario lists

COSMO-VMI:


src/utils/SortScenarioListUtils.js
```diff
@@ -0,0 +1,45 @@
...
+
+function indexOfById (list, id) {
+  for (let i = 0; i < list.length; ++i) {
+    const item = list[i];
+    if (item.id === id) {
+      return i;
+    }
+  }
+  return -1;
+}
```
csm-thu, line 18: This function can be replaced by the **find** method of arrays:
`list.find(item => item.id === id)`
If no items match the expected condition, then _undefined_ is returned.
More info here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find

csm-thu (approved):
LGTM

src/utils/SortScenarioListUtils.js
```diff
@@ -0,0 +1,45 @@
+// Copyright (c) Cosmo Tech.
+// Licensed under the MIT license.
+
+// SORT
+export const testSort = (scenarioList) => {
+  const sortedList = sortScenarioList(scenarioList.data.slice());
+  console.log(sortedList);
+};
```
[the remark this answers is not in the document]
COSMO-VMI: Nope, thanks

src/utils/SortScenarioListUtils.js
COSMO-VMI: Nice, I don't know all the these algo yet. I'll update it.

COSMO-VMI: Also, it's more `findIndex()` than `find()`

Nikeleos (approved):

src/utils/SortScenarioListUtils.js
jreynard-code: don't forget to remove the function if it's not used ;)

jreynard-code (changes requested):
LGTM, just remove the unused function ;)

PR #76, opened 2021-06-24
Download files in upload file

jreynard-code:
fix: download files in upload file
refactor:  downloadWorkspaceFile for JS client is not used anymore (fetch method is used instead for the moment)
feat: change default accepted file format for initial_stock_param (now .zip,.csv,.json,.xlsx)

### DIFF
--- a/src/services/workspace/WorkspaceService.js
+++ b/src/services/workspace/WorkspaceService.js
@@ -1,7 +1,9 @@
 // Copyright (c) Cosmo Tech.
 // Licensed under the MIT license.
 
-import { CosmotechApiService } from '../../configs/Api.config';
+import { CosmotechApiService, getDefaultBasePath } from '../../configs/Api.config';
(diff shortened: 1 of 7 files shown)


src/services/workspace/WorkspaceService.js
```diff
@@ -37,6 +39,32 @@ function getAllWorkspaceFileName (organizationId, workspaceId) {
   });
 }
 
+async function fetchWorkspaceFile (organizationId, workspaceId, filePath) {
+  const accessToken = readFromStorage(ACCESS_TOKEN_LOCAL_STORAGE_KEY);
```
csm-thu, line 43: Since we're using the Api config, maybe we could add a **getAccessToken** function there, instead of reading from local storage with ACCESS_TOKEN_LOCAL_STORAGE_KEY

csm-thu (approved):
LGTM

PR #101, opened 2021-07-26
fix: State management for upload file

jreynard-code:
Dev Review rules:
- 3 PR max

Definition of Done:
- [X] Testing
- [X] No error logs in web console (except iFrame errors)
- [] Create or Update documentation (README file for users/integrators)
- [X] Check the [test page](https://spaceport.cosmotech.com/confluence/pages/viewpage.action?spaceKey=QA&title=Azure+Sample+Webapp) in order to manually validate the feature 
- [X] Clean code (regroup configuration variable, typos, no console.log...)

### DIFF
--- a/src/components/ScenarioParameters/UploadFileUtils.js
+++ b/src/components/ScenarioParameters/UploadFileUtils.js
@@ -23,7 +23,7 @@ function buildFileNameFromDataset (dataset, storageFilePath) {
 
 // Update AZURE_STORAGE_CONTAINER_BLOB_PREFIX in a dataset reference
 function updatePathInDatasetRef (dataset, storageFilePath) {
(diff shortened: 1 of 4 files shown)


src/components/ScenarioParameters/UploadFileUtils.js
```diff
@@ -174,12 +175,46 @@ const resetUploadFile = (datasetId, file, setFile) => {
   }
 };
 
+function manageUploadFile (datasetId, file, fetchDataset, dataset, setDataset, setFile) {
```
csm-thu, line 178: Could we rename this function into "updateDatasetState" ?
Also, I'm not sure this should be a part of the UploadFIleUtils, because its only goal is to update the state of the dataset in the ScenarioParameters component

csm-thu (approved):
LGTM :+1:

PR #106, opened 2021-07-30
THU/delete_scenario_PROD-7543

csm-thu:
* create deleteScenario function in scenario service
* add redux actions, dispatchers & saga to **reset current scenario** to null (in case the last scenario is deleted) and to **delete a scenario**
* use new scenario delete feature of ScenarioManagerTreeList in the ScenarioManager view
* add labels in translation dicts


src/views/ScenarioManager/ScenarioManager.js
```diff
@@ -44,8 +63,12 @@ const ScenarioManager = (props) => {
 
 ScenarioManager.propTypes = {
   datasets: PropTypes.array.isRequired,
+  deleteScenario: PropTypes.func.isRequired,
+  findScenarioById: PropTypes.func.isRequired,
   scenarios: PropTypes.array.isRequired,
-  currentScenario: PropTypes.object.isRequired
+  resetCurrentScenario: PropTypes.func.isRequired,
+  currentScenario: PropTypes.object,
+  userId: PropTypes.string
```
jreynard-code, line 71: UserId can be required?
As scenario deletion is allowed only for scenarios that you have created, the user info is needed.

jreynard-code (changes requested):
A little change but everything LGTM ;)

src/views/ScenarioManager/ScenarioManager.js
csm-thu: Done :+1:

PR #113, opened 2021-08-05
Setup refresh token

jreynard-code:
Dev Review rules:
- 3 PR max

Definition of Done:
- [X] Testing
- [X] No error logs in web console (except iFrame errors)
- [X] Check the [test page](https://spaceport.cosmotech.com/confluence/pages/viewpage.action?spaceKey=QA&title=Azure+Sample+Webapp) in order to manually validate the feature 
- [X] Clean code (regroup configuration variable, typos, no console.log...)

_______________
Migrate from @cosmotech/api to @cosmotech/api-ts
Use axios lib for every API calls
Remove useless services
Adapt services used
Adapt sagas with new 'clientApi'
Add request interceptors for clientApi (adding Autorization header with refresh token )


src/services/dataset/DatasetService.js
```diff
@@ -1,57 +1,25 @@
...
-function findAllDatasets (organizationId) {
-  return new Promise((resolve) => {
-    DatasetApi.findAllDatasets(organizationId, (error, data, response) => {
-      resolve({ error, data, response });
-    });
-  });
-}
+import { Api } from '../../configs/Api.config';
 
 function findDatasetById (organizationId, datasetId) {
```
csm-thu, line 5: Do we really need the functions `createDataset`, `updateDataset` and `deleteDataset` ? If these functions are only wrappers around the call to the API, we should be consistent between all services and create these wrappers for all calls to the API

src/services/Client.config.js
```diff
@@ -0,0 +1,27 @@
+// Copyright (c) Cosmo Tech.
```
csm-thu, line 1: Can we discuss to find a better name than `Client.config.js` ? IMHO this is not a configuration file

src/services/Client.config.js
```diff
@@ -0,0 +1,27 @@
...
+
+import axios from 'axios';
+
+import { Auth } from '@cosmotech/core';
+
+const addInterceptors = (axiosInstance) => {
+  axiosInstance.interceptors.request.use(
+    async (request) => {
+      const tokens = await Auth.acquireTokens();
+      if (tokens !== undefined && tokens.accessToken !== undefined) {
```
csm-thu, line 12: Could be simplified in `if (tokens?.accessToken)`

csm-thu (changes requested):

src/services/Client.config.js
jreynard-code: ClientApi.js is ok for you? I named it like the constant that we export in it

csm-thu: Yup, sounds good to me :+1:

src/services/dataset/DatasetService.js
jreynard-code: By now, I let these "wrapper class" to not have a big breaking change between 1.3 and 1.4.
The ticket "Move to NPM Packages" (for v2) will normally fix this point.

src/utils/StorageUtils.js
```diff
@@ -3,6 +3,35 @@
...
+      await Auth.isUserSignedIn();
+      // Get the new token
+      authAccessToken = localStorage.getItem('authAccessToken');
+      console.log('New token');
+      console.log(authAccessToken);
+    }
+  }
+  return authAccessToken;
+}; */
+
```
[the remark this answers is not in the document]
jreynard-code: I deleted the StorageUtils file, it's not used anymore ;)

PR #159, opened 2021-09-30
Vmi/metrics

COSMO-VMI:
Dev Review rules:
- 3 PR max

Definition of Done:
- [ ] Testing

### DIFF
--- a/src/components/ScenarioParameters/ScenarioParameters.js
+++ b/src/components/ScenarioParameters/ScenarioParameters.js
@@ -244,12 +244,15 @@ const ScenarioParameters = ({
     updateAndLaunchScenario(workspaceId, scenarioId, parametersData);
     changeEditMode(false);
   };
-  const fileUploadComponent = UploadFileUtils.constructFileUpload('0',
+  const fileUploadComponent = UploadFileUtils.constructFileUpload(
+    '0',
     initialStockFile,
     setInitialStockFile,
     initialStockDataset.id,
     INITIAL_STOCK_PARAM.defaultFileTypeFilter,
-    editMode);
+    editMode,
+    scenarioId
+  );
   // Indices in this array must match indices in the tabs configuration file src/config/ScenarioParameters.js
   const scenarioParametersTabs = [
     fileUploadComponent,
--- a/src/components/ScenarioParameters/ScenarioParameters.js
+++ b/src/components/ScenarioParameters/ScenarioParameters.js
@@ -250,8 +250,7 @@ const ScenarioParameters = ({
     setInitialStockFile,
     initialStockDataset.id,
     INITIAL_STOCK_PARAM.defaultFileTypeFilter,
-    editMode,
-    scenarioId
+    editMode
   );
   // Indices in this array must match indices in the tabs configuration file src/config/ScenarioParameters.js
   const scenarioParametersTabs = [
--- a/src/components/ScenarioParameters/UploadFileUtils.js
+++ b/src/components/ScenarioParameters/UploadFileUtils.js
@@ -3,7 +3,10 @@
 
 import React from 'react';
 import { UPLOAD_FILE_STATUS_KEY } from '@cosmotech/ui';
-import { ORGANIZATION_ID, WORKSPACE_ID } from '../../config/AppInstance';
+import {
+  ORGANIZATION_ID, WORKSPACE_ID,
+  ENABLE_APPLICATION_INSIGHTS
+} from '../../config/AppInstance';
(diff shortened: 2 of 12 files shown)


src/components/ScenarioParameters/UploadFileUtils.js
```diff
@@ -3,13 +3,38 @@
...
 import { FileUpload } from './components/tabs';
+import { AppInsights } from '../../services/AppInsights';
 
 export const STORAGE_ROOT_DIR_PLACEHOLDER = '%WORKSPACE_FILE%/';
 
+// App Insigths
+const appInsights =
+  ENABLE_APPLICATION_INSIGHTS
+    ? AppInsights.getInstance()
+    : undefined;
```
csm-thu, line 21: Can we factorize the `ENABLE_APPLICATION_INSIGHTS` parameter check by putting it inside the `AppInsights` service ? If `ENABLE_APPLICATION_INSIGHTS` is false, then we could either return null or undefined when getting the singleton, or even better, prevent the service from monitoring events and metrics.

This would allow us to remove all the `if (appInsights) {...}` checks in the views and components

src/components/ScenarioParameters/ScenarioParameters.js
```diff
@@ -245,12 +245,15 @@ const ScenarioParameters = ({
...
-  const fileUploadComponent = UploadFileUtils.constructFileUpload('0',
+  const fileUploadComponent = UploadFileUtils.constructFileUpload(
+    '0',
     initialStockFile,
     setInitialStockFile,
     initialStockDataset.id,
     INITIAL_STOCK_PARAM.defaultFileTypeFilter,
-    editMode);
+    editMode,
+    scenarioId
```
csm-thu, line 255: I'm not a huge fan of carrying the `scenarioId` information in a graphical component that does not need it, only for tracking purposes. Can we store this kind of data in the `AppInsights` service singleton instead ?

For example, we could add a `setCurrentScenarioId` to the `AppInsights` service, and call this method every time the user selects a different scenario. The scenario id value could then be used in all events and metrics that need it.

src/components/ScenarioParameters/UploadFileUtils.js
```diff
@@ -3,13 +3,38 @@
...
+    appInsights.trackEvent({ name: 'UploadFile' }, { name: scenarioId });
+    appInsights.trackMetric({ name: 'UploadFileValue', average: 1, sampleCount: 1 });
+  }
+};
+  // Track download
+const trackDownload = (scenarioId) => {
+  if (appInsights) {
+    appInsights.trackEvent({ name: 'DownloadFile' }, { scenarioId: scenarioId });
+    appInsights.trackMetric({ name: 'DownloadFileValue', average: 1, sampleCount: 1 });
+  }
```
csm-thu, line 35: We should hide the implementation details of the calls to `trackEvent` and `trackMetric` in the service instead of using it in utils files and components.

Can we move `trackUpload` and `trackDownload` methods in the `AppInsights` service ?

src/services/config/ApplicationInsights.js
```diff
@@ -5,15 +5,20 @@ import { DistributedTracingModes } from '@microsoft/applicationinsights-web';
...
+        disableFetchTracking: false,
+        enableCorsCorrelation: true,
+        enableRequestHeaderTracking: true,
+        enableResponseHeaderTracking: true,
+        enableAutoRouteTracking: true,
+        distributedTracingMode: DistributedTracingModes.AI_AND_W3C
+      }
+    };
   }
+  return undefined;
```
csm-thu, line 23: This file is not supposed to have any logic in it, the check of the `APPLICATION_INSIGHTS_INSTRUMENTATION_KEY` value should be done in the service, to keep the configuration file as simple as possible.

src/views/Scenario/Scenario.js
```diff
@@ -102,6 +104,45 @@ const Scenario = (props) => {
...
+    }
+    updateAndLaunchScenario(workspaceId, scenarioId, scenarioParameters);
+  };
+
+  const launchTrackedScenario = (workspaceId, scenarioId) => {
+    if (appInsights) {
+      trackLaunchScenario();
+    }
+    launchScenario(workspaceId, scenarioId);
+  };
```
csm-thu, line 144: Could we add the tracking part directly in redux sage instead of wrapping the calls to `createScenario`, `updateAndLaunchScenario` and `launchScenario` here ?

csm-thu (changes requested):

csm-thu (approved):

PR #170, opened 2021-10-11
THU/scenario_arameters_refacto

csm-thu:


src/components/ScenarioParameters/FileManagementUtils.js
```diff
@@ -0,0 +1,215 @@
...
+
+import { UPLOAD_FILE_STATUS_KEY } from '@cosmotech/ui';
+import { ORGANIZATION_ID, WORKSPACE_ID } from '../../config/AppInstance';
+import DatasetService from '../../services/dataset/DatasetService';
+import WorkspaceService from '../../services/workspace/WorkspaceService';
+import { AppInsights } from '../../services/AppInsights';
+import { DatasetsUtils, ScenarioParametersUtils } from '../../utils';
+
+export const STORAGE_ROOT_DIR_PLACEHOLDER = '%WORKSPACE_FILE%/';
+const DATASET_ID_VARTYPE = '%DATASETID%';
```
jreynard-code, line 12: `%DATASETID%` is present in several files, it could be interesting to create a const in a config file under services/config and use this const

src/components/ScenarioParameters/components/ScenarioParametersTabs.js
```diff
@@ -40,86 +40,71 @@ const useStyles = makeStyles(theme => ({
 }));
 
 const ScenarioParametersTabs = ({
-  tabs,
+  parametersGroupsMetadata,
   currentScenario
```
jreynard-code, line 44: `currentScenario` is not used, you could remode this line (and the Props)

src/components/ScenarioParameters/components/ScenarioParametersTabs.js
```diff
@@ -40,86 +40,71 @@ const useStyles = makeStyles(theme => ({
 }));
 
 const ScenarioParametersTabs = ({
```
jreynard-code, line 42: Was thinking about, this component, with little work it could be "generalized" :thinking: right now it's not useful, maybe when we'll work on wizard component...  We'll see ;)

src/utils/scenarioParameters/ScenarioParameterInputFactory.js
```diff
@@ -0,0 +1,41 @@
+// Copyright (c) Cosmo Tech.
```
jreynard-code, line 1: Maybe rename folder with 'factory' in names at least (same thing for 'inputComponent')

src/components/ScenarioParameters/components/ScenarioParametersTabs.js
csm-thu: Yes, this component defines the layout of the scenario parameters tabs, we can imagine adding support for more layouts in the future.

src/utils/scenarioParameters/inputComponents/BasicNumberInputFactory.js
```diff
@@ -0,0 +1,56 @@
+// Copyright (c) Cosmo Tech.
+// Licensed under the MIT license.
+
+import React from 'react';
+import { BasicNumberInput } from '@cosmotech/ui';
+
+const DEFAULT_MIN_VALUE = -1e10 + 1;
```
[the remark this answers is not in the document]
csm-thu: I preferred limiting the number of digits to 10 instead of using a language limit value that would seem random to users :)

src/utils/DatasetsUtils.js
```diff
@@ -0,0 +1,51 @@
...
+
+// Retrieve dataset file location in Azure Storage
+function getStorageFilePathFromDataset (dataset) {
+  const blobPrefix = dataset?.connector?.parametersValues?.AZURE_STORAGE_CONTAINER_BLOB_PREFIX;
+  if (blobPrefix !== undefined) {
+    return blobPrefix.split(STORAGE_ROOT_DIR).pop();
+  }
+}
+
+// Retrieve file name from dataset information
```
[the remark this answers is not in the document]
csm-thu: If we must support custom connector types, I'd rather have them in a distinct file to prevent projects cloning the webapp from getting conflicts when we make changes in the utils files, but we should save it for later indeed :)

Nikeleos (commented):
Some comments, especially in Cypress parts.

cypress/commons/actions/generic/Scenarios.js
```diff
@@ -0,0 +1,56 @@
...
+}
+function getScenarioCreationDialogRunTypeSelector () {
+  return cy.get(SELECTORS.scenario.createDialog.typeSelect);
+}
+function getScenarioCreationDialogSubmitButton () {
+  return cy.get(SELECTORS.scenario.createDialog.submitButton);
+}
+
+// From scenario view, select the scenario with the provided name and id
+function select (scenarioName, scenarioId) {
```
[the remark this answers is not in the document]
csm-thu: Actually the short name was set on purpose, because it is part of a "Scenarios" module, and will then be called with `Scenarios.select(...)`.
`Scenarios.selectScenario(...)` would be redundant I think

PR #185, opened 2021-10-19
THU/refactor_brewery_selectors_in_cypress_tests

csm-thu:
Separate generic data from data specific to brewery in cypress folder


jreynard-code (approved):
LGTM ;)

PR #250, opened 2022-01-20
fix: fix empty field when typing zero in a number input

csm-thu:
It was also impossible to enter values between -1 and 1,
these values were rounded to zero and fell into the same bug.

### DIFF
--- a/src/utils/scenarioParameters/factories/inputComponentsFactories/BasicNumberInputFactory.js
+++ b/src/utils/scenarioParameters/factories/inputComponentsFactories/BasicNumberInputFactory.js
@@ -38,12 +38,17 @@ const create = (t, parameterData, parametersState, setParametersState, editMode)
     });
   }
 
+  let value = parametersState[parameterData.id];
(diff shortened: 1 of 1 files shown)


src/utils/scenarioParameters/factories/inputComponentsFactories/BasicNumberInputFactory.js
```diff
@@ -38,12 +38,17 @@ const create = (t, parameterData, parametersState, setParametersState, editMode)
     });
   }
 
+  let value = parametersState[parameterData.id];
+  if (value === undefined || value === null) {
```
nborde-CSM, line 42: Sorry, I have been a little too quicly for the review and after watching it closer, I noticed that :
 
 (value === undefined || value === null)  is the same thing than (value == null)

csm-thu: Replaced the check by `value == null` :+1:

PR #253, opened 2022-01-27
Nbo/default scenario shown

nborde-CSM:
Dev Review rules:
- 3 PR max

Definition of Done:
- [ ] Testing


src/utils/SortScenarioListUtils.js
```diff
@@ -27,3 +27,15 @@ export const sortScenarioList = (scenarioList) => {
 
   return sortedList;
 };
+
+export const getFirstScenarioMaster = (scenarioList) => {
+  if (scenarioList.length > 0) {
+    const scenarioMasterList = scenarioList.filter((scenario) => scenario.parentId === null);
+
+    return scenarioMasterList.sort((scenarioA, scenarioB) =>
+      scenarioB.name.toUpperCase() > scenarioA.name.toUpperCase() ? -1 : 1
```
csm-thu, line 36: In the function just above this one, we are using
`a.name.toUpperCase() < b.name.toUpperCase()` instead of `scenarioB.name.toUpperCase() > scenarioA.name.toUpperCase` to sort the scenarios alphabetically (note that the comparison operator is different).
Could you factorize this in a separate function that we can reuse please ? (this would prevent errors such as different comparison operators in the sorting function)

csm-thu (changes requested):

csm-thu (approved):
LGTM :+1:

src/utils/SortScenarioListUtils.js
```diff
@@ -27,3 +33,12 @@ export const sortScenarioList = (scenarioList) => {
 
   return sortedList;
 };
+
+export const getFirstScenarioMaster = (scenarioList) => {
+  if (scenarioList.length > 0) {
+    const scenarioMasterList = scenarioList.filter((scenario) => scenario.parentId === null);
+    return _sortByName(scenarioMasterList)[0];
```
csm-thu, line 40: If the scenario list data is corrupted, it may contain only non-root scenarios (scenarios whose parent is not null). In this case, `scenarioMasterList` will be empty and reading its first element with `_sortByName(scenarioMasterList)[0]` will make the application crash.
Could we add the following safeguard please ?
```
const scenarioMasterList = scenarioList.filter((scenario) => scenario.parentId === null);
if (scenarioMasterList.length === 0) {
  console.warn('Scenario list is corrupted, no root scenarios found');
  return _sortByName(scenarioList)[0];
}
return _sortByName(scenarioMasterList)[0];
```

csm-thu (approved):
One addition, but otherwise LGTM :+1:

PR #274, opened 2022-03-18
fix: fix login error on redirection with some browsers

csm-thu:
BREAKING CHANGE: the redirection URL was changed from /scenario to /sign-in. Solutions
must update the accepted redirection URLs in the Azure portal of the App Registration
and replace /scenario by /sign-in (or only add /sign-in to keep supporting previous
versions of their webapps)


csm-thu:
Commit has been cherry-picked on branch **release/3.0.0**, closing this PR

PR #292, opened 2022-04-15
Nbo/make cypress tests more robust

nborde-CSM:
Dev Review rules:
- 3 PR max

Definition of Done:
- [ ] Testing
- [ ] Identify breaking changes with [conventional commits](https://kapeli.com/cheat_sheets/Conventional_Commits.docset/Contents/Resources/Documents/index)


csm-thu:
Could you change the prefix in commit [fix: add keystrokeDelay to 0 in scenario launch tests](https://github.com/Cosmo-Tech/azure-sample-webapp/pull/292/commits/ec903ad96b8930377cc985d24129f543e0c75e71), from `fix` to `test`, please ?

cypress/commons/actions/generic/ScenarioParameters.js
```diff
@@ -62,7 +62,7 @@ function collapseParametersAccordion() {
       }
     });
 }
-function edit(timeout = 5) {
+function edit(timeout) {
```
csm-thu, line 65: Can we leave a default value in this function when we want to edit without waiting the end of a scenario run, or does it make the tests break ?

cypress/commons/actions/generic/Scenarios.js
```diff
@@ -51,15 +52,25 @@ function switchToScenarioView() {
 }
 
 // Select the scenario with the provided name and id
-function selectScenario(scenarioName, scenarioId) {
+function selectScenario(scenarioName, scenarioId, selectTimeout) {
```
csm-thu, line 55: Can we have a default value for the `selectTimeout` parameter ? This argument does not seem to be mandatory when writing the tests, and we want to keep the tests body as simple as possible.

cypress/integration/brewery/FileParameters-tests.spec.js
```diff
@@ -117,10 +117,8 @@ describe('Simple operations on a file parameter', () => {
 
   it('can delete an uploaded file and run the scenario', () => {
     // Wait required to prevent some kind of race condition somehow
-    cy.wait(5000); // eslint-disable-line cypress/no-unnecessary-waiting
-    Scenarios.selectScenario(firstScenarioName, firstScenarioId);
-    cy.wait(5000); // eslint-disable-line cypress/no-unnecessary-waiting
-    ScenarioParameters.edit(120);
+    Scenarios.selectScenario(firstScenarioName, firstScenarioId, 5);
```
csm-thu, line 120: `5` could be the default value for the `selectTimeout` argument (c.f. my previous comment)

cypress/integration/brewery/ScenarioLaunch-test.spec.js
```diff
@@ -15,6 +15,10 @@ describe('Create scenario', () => {
     scenariosNames[i] = SCENARIO_NAME.SCENARIO_MASTER + i + ' - ' + randomString;
   }
 
+  Cypress.Keyboard.defaults({
+    keystrokeDelay: 0,
+  });
+
```
csm-thu, line 21: We may need this change of keystrokeDelay value for the whole test file, can we move these lines at the top of the file, right after the imports ? (like we did in test files for Table and File parameters)

csm-thu (changes requested):

cypress/commons/actions/generic/ScenarioParameters.js
nborde-CSM: In fact, a default value is already implemented at the lower level directly in the getter function :
`function getParametersEditButton(timeout = 4) {
  return cy.get(GENERIC_SELECTORS.scenario.parameters.editButton, { timeout: timeout * 1000 });
}`

csm-thu: Even though a default value is set in `getParametersEditButton`, if we remove the default value in the `edit` function, **we can no longer call `edit` without parameter in the tests files**.
That means that even if we want to edit parameters without waiting the end of a scenario run, we have to set explicitly a waiting time -> some tests file could be simpler and easier to read without the timeout value

cypress/integration/brewery/FileParameters-tests.spec.js
nborde-CSM: Same as above, I put default timeout in getter function :
`function getScenarioSelector(timeout = 5) {
  return cy.get(GENERIC_SELECTORS.scenario.selectInput, { timeout: timeout * 1000 });
}`

cypress/commons/actions/generic/Scenarios.js
nborde-CSM: Same as above, I put default timeout in getter function :
`function getScenarioSelector(timeout = 5) {
  return cy.get(GENERIC_SELECTORS.scenario.selectInput, { timeout: timeout * 1000 });
}`

csm-thu (approved):
LGTM, don't forget to reword the "fix" commit :+1:

PR #339, opened 2022-06-09
THU/improve_file_parameters_management

csm-thu:
- use state setters with function form to update state based on its previous value
- make dependencies arrays more specific in some `useEffect`s to prevent unnecessary refresh
- minor changes in table factory to fix missing loading spinner


esasova (approved):
LGTM :+1:

PR #343, opened 2022-06-14
Esas/runtype dependents scenario report

esasova:
Change dashboard config file (src/config/Dashboard.js) :
replace SCENARIO_DASHBOARD_CONFIG with SCENARIO_VIEW_REPORTS_BY_RUNTEMPLATE.
The new config contains all possible dashboards for the project ; every dashboard contains possible filters and other options and is identified with a key that corresponds to currentScenario.data.runTemplateId in Redux. If the same config is used for several runTemplates, it must be declared for each runTemplate (in the defaultScenarioReport variable for example). 

Then, in Scenario view (/src/views/Scenario/Scenario.js), the config is stored in the currentScenarioRunTemplateReport array that helps distinguish new config from old one (for backward compatibility currentScenarioRunTemplateReport must be an array to be compatible with existing SimplePowerBIReportEmbed component(cosmotech/ui library)) : if the old one, the entire config object is assigned to currentScenarioRunTemplateReport, if new one, the SCENARIO_VIEW_REPORTS_BY_RUNTEMPLATE array is filtered for runTemplateId and the result is pushed in currentScenarioViewReport. To avoid app crash when there is no scenario, the first report in the list is declared as defaultPowerBIReport.

The powerbi-client-react actually uses setFilters() and removeFilters() methods to update the configuration while Microsoft recommends the general updateFilters() method : https://docs.microsoft.com/en-us/javascript/api/overview/powerbi/control-report-filters#update-filters
  This causes an undefined error in console nevertheless the config is updated properly. To avoid any confusion or future incompatibility, the embed component is destroyed and rerendered when a scenario changes.

Dev Review rules:
- 3 PR max

Definition of Done:
- [ ] Testing
- [ ] Identify breaking changes with [conventional commits](https://kapeli.com/cheat_sheets/Conventional_Commits.docset/Contents/Resources/Documents/index)

### DIFF
--- a/src/config/Dashboards.js
+++ b/src/config/Dashboards.js
@@ -29,7 +29,7 @@ const defaultScenarioViewReport = {
       },
     },
   },
-  staticFilters: [new PowerBIReportEmbedMultipleFilter('Bar', 'Bar', ['MyBar'])],
+  staticFilters: [new PowerBIReportEmbedMultipleFilter('Bar', 'Bar', ['MyBar', 'MyBar2'])],
   dynamicFilters: [
     new PowerBIReportEmbedSimpleFilter('StockProbe', 'SimulationRun', POWER_BI_FIELD_ENUM.SCENARIO_CSM_SIMULATION_RUN),
     new PowerBIReportEmbedSimpleFilter('Bar', 'simulationrun', POWER_BI_FIELD_ENUM.SCENARIO_CSM_SIMULATION_RUN),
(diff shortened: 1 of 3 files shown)


nborde-CSM (approved):
All LGTM :+1:

src/views/Scenario/Scenario.js
```diff
@@ -323,8 +328,9 @@ const Scenario = (props) => {
       </Grid>
       <Card>
         <SimplePowerBIReportEmbed
+          key={currentScenario?.data?.id}
```
csm-thu, line 331: Do we need to add a key to this component ?

src/config/Dashboards.js
```diff
@@ -13,37 +13,86 @@ import {
...
       panes: {
         filters: {
-          expanded: false,
-          visible: false,
+          expanded: true,
+          visible: true,
         },
       },
     },
-    staticFilters: [new PowerBIReportEmbedMultipleFilter('Bar', 'id', ['MyBar', 'MyBar2'])],
```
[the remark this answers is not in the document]
csm-thu: Ok, thank you :)

src/views/Scenario/Scenario.js
csm-thu: Ok, adding a comment to explain why we should keep the `key` might be helpful later :smile:

csm-thu (approved):
LGTM :+1:

PR #346, opened 2022-06-21
THU/brewery_cytoviz_view

csm-thu:


public/locales/en/translation.json
```diff
@@ -91,6 +91,30 @@
...
+      "elementData": {
+        "dictKey": "Key",
+        "dictValue": "Value",
+        "noData": "No data to display for this element.",
+        "attributes": {
+          "PlantName": "Plant"
+        }
+      },
+      "placeholder": {
+        "error": "An error occured, can not visualize data.",
```
esasova, line 114: famous can not VS cannot :smile:

src/utils/ErrorsUtils.js
```diff
@@ -0,0 +1,44 @@
...
+  } else if (error?.request) {
+    return {
+      title: error.request.statusText,
+      status: error.request.status,
+      detail: error.request.response,
+    };
+  }
+
+  return {
+    title: 'Unexpected error',
```
esasova, line 40: While creating the ErrorBanner component, we discussed with Emma the "unexpected error" and we decided to opt for "unknown error" because an error is never expected :smile: should we follow the same guideline here, what do you think ?

src/views/Instance/Instance.js
```diff
@@ -0,0 +1,172 @@
...
+  if (noScenario) {
+    cytoVizPlaceholderMessage = t(
+      'commoncomponents.cytoviz.placeholder.noScenario',
+      'No scenario. You can create a new scenario in the Scenario view.'
+    );
+  }
+  if (errorBannerMessage) {
+    cytoVizPlaceholderMessage = t(
+      'commoncomponents.cytoviz.placeholder.error',
+      'An error occured, can not visualize data.'
```
esasova, line 129: cannot :smile:

src/views/Instance/data.js
```diff
@@ -0,0 +1,146 @@
...
+  const { style, ...otherMetadata } = edgesGroupMetadata;
+  return {
+    group: 'edges',
+    data: edge,
+    classes: classes,
+    ...otherMetadata,
+  };
+};
+
+export const _processGraphCompounds = (dataset, compoundsGroups) => {
```
esasova, line 41: is export necessary here if this function isn't used elsewhere ?

src/views/Instance/data.js
```diff
@@ -0,0 +1,146 @@
...
+};
+
+async function _fetchDataFromADT(scenarioId, dataSource) {
+  const tokens = await Auth.acquireTokens();
+  const headers = { 'x-functions-key': dataSource.functionKey };
+  if (tokens?.accessToken) {
+    headers.common = {};
+    headers.common.Authorization = 'Bearer ' + tokens.accessToken;
+  }
+  return await axios({
```
esasova, line 116: I've got an eslint warning about `return await`, maybe we can disable the rule if we need it ? [https://eslint.org/docs/rules/no-return-await](https://eslint.org/docs/rules/no-return-await)

esasova (approved):
LGTM :+1:  just a couple of questions :slightly_smiling_face:

src/utils/ErrorsUtils.js
csm-thu: Good idea, thanks for the suggestion :)

src/views/Instance/data.js
csm-thu: Nice catch, classic copy/paste mistake :smile:

src/views/Instance/Instance.js
```diff
@@ -0,0 +1,172 @@
...
+const EXTRA_LAYOUTS = {
+  breadthfirst: null,
+};
+const appInsights = AppInsights.getInstance();
+
+const Instance = (props) => {
+  const classes = useStyles();
+  const { t } = useTranslation();
+  const { currentScenario, scenarioList, findScenarioById, workspace } = props;
+  const [graphElements, setGraphElements] = useState([]);
```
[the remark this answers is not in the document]
csm-thu: Not sure what you would suggest instead, but the `CytoViz` component defines `elements: PropTypes.array.isRequired,`, so a null/undefined value would not be accepted as prop

src/views/Instance/data.js
```diff
@@ -0,0 +1,146 @@
...
+} from './styleCytoViz';
+import { ORGANIZATION_ID, WORKSPACE_ID } from '../../config/AppInstance';
+import instanceViewData from '../../config/InstanceVisualization.js';
+
+const IS_INSTANCE_VIEW_FUNCTION_CONFIG_VALID = !(instanceViewData.dataSource == null);
+export { IS_INSTANCE_VIEW_FUNCTION_CONFIG_VALID };
+
+const _formatLabelWithNewlines = (label) => label?.replace(/[_|\s]/g, '\n') || '';
+
+const _forgeCytoscapeNodeData = (node, classes, nodesGroupMetadata = {}, nodesParentsDict = {}) => {
```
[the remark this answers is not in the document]
csm-thu: IMO, it makes the code simpler to write because we don't have to check whether these parameters are null or not:
```
  const parent = nodesParentsDict[node.id];
  const { style, ...otherMetadata } = nodesGroupMetadata;
```

src/views/Instance/data.js
```diff
@@ -0,0 +1,146 @@
...
+      style: { ...DEFAULT_NODE_STYLE, ...nodesGroupMetadata.style },
+    });
+    processedData.stylesheet.push({
+      selector: `node.${nodesGroupName}:selected`,
+      style: { ...DEFAULT_SELECTED_NODE_STYLE, ...nodesGroupMetadata.style },
+    });
+  });
+};
+
+export const _processGraphEdges = (processedData, dataset, edgesGroups) => {
```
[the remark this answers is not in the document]
csm-thu: Fixed with other occurences, thank you :)

src/views/Instance/Instance.js
```diff
@@ -0,0 +1,172 @@
...
+
+  return (
+    <>
+      <Backdrop className={classes.backdrop} open={showBackdrop}>
+        <CircularProgress color="inherit" />
+      </Backdrop>
+      <div className={classes.mainGrid}>
+        <div className={classes.scenarioSelectGridItem}>
+          <HierarchicalComboBox
+            value={currentScenario?.data}
```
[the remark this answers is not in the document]
csm-thu: Done :)

PR #348, opened 2022-06-24
chore: create vanilla config files

esasova:
Create a vanilla configuration file for every config :
- string values became empty strings
- object and array values became empty objects and arrays
- number and boolean values are considered as default and still the same as in brewery solution
- some values, e.g., SUPPORT_URL still the same

### DIFF
--- a/src/config/AppConfiguration.vanilla.js
+++ b/src/config/AppConfiguration.vanilla.js
@@ -1,0 +1,55 @@
+// Copyright (c) Cosmo Tech.
+// Licensed under the MIT license.
+
+// To configure your app see:
+// https://github.com/Cosmo-Tech/azure-sample-webapp/blob/main/doc/appConfiguration.md
+
+// App version
+export const APP_VERSION = process.env.REACT_APP_VERSION;
+
+// App translation configuration
+export const LANGUAGES = {
+  en: 'English',
+  fr: 'Français',
+};
+
+export const FALLBACK_LANGUAGE = 'en'; // Language used if detection fails
+
+// Duration of inactivity after which users are automatically logged out (minutes)
+export const SESSION_INACTIVITY_TIMEOUT = 30;
+
+// Polling delay to update running scenario status (milliseconds)
+export const SCENARIO_STATUS_POLLING_DELAY = 10000;
+
+// Polling delay to update Power BI information (milliseconds)
+export const POWER_BI_INFO_POLLING_DELAY = 10000;
+
+// Type of logs to download for a scenario run (SIMPLE_LOGS or CUMULATED_LOGS)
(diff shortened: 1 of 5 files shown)


src/config/AppConfiguration.vanilla.js
```diff
@@ -0,0 +1,52 @@
...
+// To configure your app see:
+// https://github.com/Cosmo-Tech/azure-sample-webapp/blob/main/doc/appConfiguration.md
+
+// App version
+export const APP_VERSION = process.env.REACT_APP_VERSION;
+
+// App translation configuration
+export const LANGUAGES = {};
+
+export const FALLBACK_LANGUAGE = ''; // Language used if detection fails
```
csm-thu, line 13: We could keep the values defined for both languages currently supported by the sample webapp:
```
// App translation configuration
export const LANGUAGES = {
  en: 'English',
  fr: 'Français',
};
export const FALLBACK_LANGUAGE = 'en'; // Language used if detection fails
```

The goal of the "vanilla" version is to remove the elements that are specific to the brewery solution, but we should keep the parts that are "common" to all solutions and can be used as default values when creating a new webapp (e.g. the "en" and "fr" languages that are already supported).

src/config/AppConfiguration.vanilla.js
```diff
@@ -0,0 +1,52 @@
...
+export const SESSION_INACTIVITY_TIMEOUT = 30;
+
+// Polling delay to update running scenario status (milliseconds)
+export const SCENARIO_STATUS_POLLING_DELAY = 10000;
+
+// Polling delay to update Power BI information (milliseconds)
+export const POWER_BI_INFO_POLLING_DELAY = 10000;
+
+// Type of logs to download for a scenario run (SIMPLE_LOGS or CUMULATED_LOGS)
+export const SCENARIO_RUN_LOG_TYPE = '';
```
csm-thu, line 25: In order to make the first configuration easier for solutions, we can use the same value as the one we're using in the sample webapp for this parameter: new solutions can use "CUMULATED_LOGS" by default.
`export const SCENARIO_RUN_LOG_TYPE = 'CUMULATED_LOGS';`

src/config/AppConfiguration.vanilla.js
```diff
@@ -0,0 +1,52 @@
...
+
+// Whether to enable Application Insights events tracking
+export const ENABLE_APPLICATION_INSIGHTS = true;
```
csm-thu, line 28: This might raise some issues related to user-data collection and would require a correct value for the AppInsights instrumentation key. Could you check with @FMO-Cosmo ? Maybe we should set this parameter to `false` by default :slightly_smiling_face:

csm-thu (approved):
LGTM :+1:

PR #384, opened 2022-08-10
Thu/improve cy tests robustness

csm-thu:
- fix "page load" cypress error related to cross-origin frame
- increased timeout value when waiting for the scenario run request
- refactored functions related to run template selector in scenari ocreation dialog


esasova (approved):
LGTM :+1:

PR #386, opened 2022-08-11
feat: compound style, desciption new switch en/fr

OskarHowe:
- new compound (not)collapsed style: compound nodes have dashed borders
- 2 selectors to apply these styles
- label in fr/en for new Switch Component in CytoViz

Attention: used lib has a bug when changing the layout in combination with clustered nodes: [Issue](https://github.com/iVis-at-Bilkent/cytoscape.js-expand-collapse/issues/138)

### DIFF
--- a/src/views/Instance/Instance.js
+++ b/src/views/Instance/Instance.js
@@ -93,6 +93,7 @@ const Instance = (props) => {
       title: t('commoncomponents.cytoviz.settings.title', 'Settings'),
       spacingFactor: t('commoncomponents.cytoviz.settings.spacingFactor', 'Spacing factor'),
       zoomLimits: t('commoncomponents.cytoviz.settings.zoomLimits', 'Min & max zoom'),
+      collapse: t('commoncomponents.cytoviz.settings.collapse', 'Compound nodes collapsed (Right click to open)'),
     },
     elementData: {
       dictKey: t('commoncomponents.cytoviz.elementData.dictKey', 'Key'),
@@ -118,6 +119,7 @@ const Instance = (props) => {
     maxZoom: 1,
     useCompactMode: true,
     spacingFactor: 1,
+    collapseAllCompounds: true,
   };
 
   let cytoVizPlaceholderMessage = null;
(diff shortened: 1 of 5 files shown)


src/views/Instance/styleCytoViz.js
```diff
@@ -43,3 +43,14 @@ export const getDefaultSelectedNodeStyle = (theme) => ({
...
 });
+
+export const getDefaultCompoundNodeStyle = (theme) => ({
+  'border-style': 'dashed',
+  'border-color': EDGE_DEFAULT_COLOR,
+  'border-width': 3,
+});
+export const getDefaultCollapsedCompoundNodeStyle = (theme) => ({
+  'border-style': 'solid',
+  'border-color': EDGE_DEFAULT_COLOR,
```
csm-thu, line 54: We should probably define a new constant `COMPOUND_DEFAULT_COLOR` instead of using `EDGE_DEFAULT_COLOR` here :smile:

src/views/Instance/Instance.js
```diff
@@ -118,6 +119,7 @@ const Instance = (props) => {
     maxZoom: 1,
     useCompactMode: true,
     spacingFactor: 1,
+    collapseAll: true,
```
csm-thu, line 122: Could we rename the prop `collapseAll` to `collapseAllCompounds` ?

src/views/Instance/data.js
```diff
@@ -69,6 +71,14 @@ const _processGraphNodes = (processedData, nodesParentsDict, datasetContent, nod
       selector: `node.${nodesGroupName}:selected`,
       style: { ...getDefaultSelectedNodeStyle(theme), ...nodesGroupMetadata.style },
     });
+    processedData.stylesheet.push({
+      selector: `node:parent`,
+      style: { ...getDefaultCompoundNodeStyle(theme) },
+    });
```
csm-thu, line 77: Can integrators still define custom styles for compound nodes (e.g. "Bar" entities) in `DATA_CONTENT` in _src/config/InstanceVisualization.js_ ?

csm-thu:
Feature discontinued in v7.0 of the webapp, closing the PR

PR #397, opened 2022-08-29
THU/fix_scenario_switch_backdrop

csm-thu:
- fix z-index of backdrop on scenario creation
- fix backdrop that wasn't displayed anymore when switching between scenarios
- refactor Instance view to prevent two loading spinners from being displayed: use `isLoading` prop of CytoViz component instead of adding another Backdrop in the view

### DIFF
--- a/src/views/Scenario/Scenario.js
+++ b/src/views/Scenario/Scenario.js
@@ -301,7 +301,7 @@ const Scenario = (props) => {
   };
   return (
     <>
-      <Backdrop open={showBackdrop}>
+      <Backdrop open={showBackdrop} style={{ zIndex: '10000' }}>
         <CircularProgress color="inherit" />
       </Backdrop>
       <div data-cy="scenario-view" className={classes.content}>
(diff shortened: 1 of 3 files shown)


src/views/Scenario/Scenario.js
```diff
@@ -301,7 +301,7 @@ const Scenario = (props) => {
   };
   return (
     <>
-      <Backdrop open={showBackdrop}>
+      <Backdrop open={showBackdrop} style={{ zIndex: '10000' }}>
```
COSMO-EMI, line 304: I don't like it. Maybe add a comment saying "Hey, this is ugly, please fix me better."

COSMO-EMI (changes requested):
Just one thing about the z-index, otherwise LGTM.

src/views/Scenario/Scenario.js
csm-thu: There aren't many ways to fix it, the only improvement we can do right now is extract the style and use a className, would it be ok for you ?
Note: even with mui v5, I can't find any `z-index` defined in the [Backdrop component](https://github.com/mui/material-ui/blob/master/packages/mui-material/src/Backdrop/Backdrop.js), so we may have to use the `sx` prop to define a custom style anyway...

COSMO-EMI: OK, it seems MUI, for some good reason, doesn't define `z-index` for Backdrop like other ones.
It's hard to tell if we should specify it each time, or globally, in an override file.

For the time being, I accept this fix :)

COSMO-EMI:
LGTM

COSMO-EMI (approved):
LGTM

esasova (approved):

PR #408, opened 2022-09-07
refactor: remove unnecessary call to back-end

csm-thu:
When opening the webapp, after receiving the list of all scenarios, we don't need to call `fetchScenarioByIdForInitialData` to write the current scenario data in redux :slightly_smiling_face:


ardvi (approved):

esasova (approved):
LGTM :+1:

PR #418, opened 2022-09-22
test: add stubbing to tests without launch or update scenario

esasova:
- [ ] Code is clean

### DIFF
--- a/cypress/commons/actions/generic/Login.js
+++ b/cypress/commons/actions/generic/Login.js
@@ -11,16 +11,11 @@ function getMicrosoftLoginButton() {
   return cy.get(GENERIC_SELECTORS.login.microsoftLoginButton);
 }
 
-function login() {
+function login(url = BASE_URL) {
   const reqAuthAlias = api.interceptAuthentication();
-  const reqPowerBIAlias = api.interceptPowerBIAzureFunction();
-  const reqGetScenariosAlias = api.interceptGetScenarios();
-  const reqGetDatasetsAlias = api.interceptGetDatasets();
-  const reqGetWorkspaceAlias = api.interceptGetWorkspace();
-  const reqGetSolutionAlias = api.interceptGetSolution();
-
+  const newPageQueries = api.interceptNewPageQueries();
   cy.clearLocalStorageSnapshot();
-  cy.visit(BASE_URL, {
+  cy.visit(url, {
     // next line defines English as default language for tests
     onBeforeLoad(win) {
       Object.defineProperty(win.navigator, 'languages', {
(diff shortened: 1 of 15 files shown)


cypress/commons/constants/generic/TestConstants.js
```diff
@@ -45,3 +45,8 @@ export const API_REGEX = {
 };
 
 export const SCENARIO_RUN_IN_PROGRESS = 'Scenario run in progress...';
+
+export const WEBAPP_URL_REGEX = {
+  SCENARIO_PAGE: new RegExp(`/${PAGE_NAME.SCENARIO}`),
+  SCENARIO_PAGE_WITH_ID: new RegExp(`${PAGE_NAME.SCENARIO}/.*`),
```
csm-thu, line 51: Do we want the SCENARIO_PAGE regex to only match URLs without scenario ids ? If yes, then the regex should end with `$` to make sure it does not match URLs with scenario ids.
Also, we could use a "group capture" in the SCENARIO_PAGE_WITH_ID regex instead of .* (e.g. `${PAGE_NAME.SCENARIO}/(s-[\\w]+)` )

cypress/integration/brewery/Router - wrong url.spec.js
```diff
@@ -1,14 +1,27 @@
...
+      GET_DATASETS: true,
+      GET_SCENARIOS: true,
+    });
+  });
+  after(() => {
+    stub.stop();
+  });
   it("can display error banner when scenario doesn't exist", () => {
     cy.clearLocalStorageSnapshot();
     cy.visit('scenario/invalidurl');
```
csm-thu, line 19: When the webapp finds out that the scenario does not exist, does it trigger a findScenarioById request on the first scenario of the scenarios list ? If it is the case maybe we should stub this request too to prevent a call to the back-end (especially if GET_SCENARIOS is enabled in the stubbing, we well query the back-end for a non-existing scenario)

csm-thu (approved):
Some remarks, but otherwise LGTM :+1:

ardvi (approved):

cypress/integration/brewery/Router - wrong url.spec.js
esasova: No, the webapp doen't trigger a findScenarioById, I think it's because we fired this API call and currentScenario is chosen from getAllScenarios list (and this list is stubbed). I tried to stubb the findScenarioById for invalidurl (to search the scenario in the fake workspace) but with no success so this is a real API call. I'll search better today :slightly_smiling_face:

cypress/commons/constants/generic/TestConstants.js
esasova: Great, thank you ! :slightly_smiling_face:

cypress/commons/actions/generic/Login.js
```diff
@@ -29,25 +24,17 @@ function login() {
...
-  api.waitAlias(reqGetDatasetsAlias, { timeout: 60 * 1000 });
-  api.waitAlias(reqGetWorkspaceAlias, { timeout: 60 * 1000 });
-  api.waitAlias(reqGetSolutionAlias, { timeout: 60 * 1000 });
-  api.waitAlias(reqPowerBIAlias);
+  api.waitNewPageQueries(newPageQueries);
   Scenarios.getScenarioViewTab(60).should('be.visible');
   cy.saveLocalStorage();
 }
 
 function relogin() {
```
csm-thu, line 33: Could we add a `url` parameter for `relogin` too ?

csm-thu (approved):
One minor suggestion, but LGTM :+1:

cypress/commons/actions/generic/Login.js
esasova: Good idea !

PR #495, opened 2022-12-21
THU/instance_viz_config_refacto

csm-thu:


src/views/Instance/Instance.js
```diff
@@ -64,10 +71,14 @@ const Instance = (props) => {
         setIsLoadingData(false);
       } else {
         try {
-          const scenario = await fetchData(organizationId, workspaceId, currentScenario.data?.id);
+          const scenario = await fetchData(instanceViewConfig, organizationId, workspaceId, currentScenario.data?.id);
           // TODO: (refactor) to improve performance, we don't need to recompute the whole graph elements set when the
           // theme is changed, we could rebuild only the stylesheet
-          const { graphElements: newGraphElements, stylesheet } = await processGraphElements(scenario.data, theme);
+          const { graphElements: newGraphElements, stylesheet } = await processGraphElements(
```
esasova, line 77: I'm not sure this is important but I have a warning for this line:
Redundant 'await' for a non-promise type
but it can be just a glitch of WebStorm :slightly_smiling_face:

esasova (approved):
LGTM :+1:

src/views/Instance/Instance.js
csm-thu: The function called with `await` is not an `async` function, thank you for pointing this out :smile:

ardvi (approved):
LGTM

PR #536, opened 2023-02-15
Thu/xlsx export prod 8222

csm-thu:
- merge target is branch **_NBO/migrate_to_MUI_5_PROD-8916_** to use mui v5 directly
- new feature: in edtiable table, "Export to CSV" button is now "Export", a pop-up lets users choose the export type (CSV or Excel) and the filename before exporting it
- cypress tests:
  - export empty table to XSLX (header must still be visible in the exported file)
  - export non-empty table export to XLSX
  - export table to XLSX & CSV with custom file names
  - existing tests with CSV exports adapted to new UI


ardvi (approved):
LGTM

src/components/ScenarioParameters/components/ScenarioParametersInputs/components/TableExportDialog/TableExportDialog.js
```diff
@@ -0,0 +1,133 @@
...
+
+TableExportDialog.propTypes = {
+  defaultFileName: PropTypes.string,
+  labels: PropTypes.shape({
+    cancel: PropTypes.string.isRequired,
+    export: PropTypes.string.isRequired,
+    fileNameInputLabel: PropTypes.string.isRequired,
+    fileTypeSelectLabel: PropTypes.string.isRequired,
+    title: PropTypes.string.isRequired,
+    label: PropTypes.string.isRequired,
```
esasova, line 116: Is it necessary to make these labels required ? They seem quite generic to me :slightly_smiling_face:

src/components/ScenarioParameters/components/ScenarioParametersInputs/components/TableExportDialog/TableExportDialog.js
```diff
@@ -0,0 +1,133 @@
...
+  const confirmExport = () => {
+    const exportedFileName = `${fileName}.${fileType}`;
+    onExport(exportedFileName);
+  };
+
+  return (
+    <Dialog onClose={onClose} open={open} data-cy="table-export-dialog">
+      <DialogTitle>{labels.title}</DialogTitle>
+      <DialogContent sx={{ paddingBottom: '0' }}>
+        <Typography>{labels.label}</Typography>
```
esasova, line 54: Maybe we can find a more explicit name for this element like _subtitle_ or something else ? :slightly_smiling_face:

esasova (approved):
LGTM :+1:

src/components/ScenarioParameters/components/ScenarioParametersInputs/components/TableExportDialog/TableExportDialog.js
csm-thu: I'd say this is necessary to inform developers that when providing the `labels` prop, all these keys must be present, otherwise some labels will be missing. IIRC, this behavior comes from the way React `defaultProps` works:
- if the prop `labels` is undefined, then its value will be set to the one of `TableExportDialog.defaultProps.labels` (with all default labels we defined)
- if the prop `labels` is defined, but some keys are missing, they won't be copied from `TableExportDialog.defaultProps.labels` leading to some text being missing

A work-around for this would be to "patch" the `labels` prop received with a dict containing all default values, but the result is less readable, and I'm not sure it is very conventional :sweat_smile: (see [this PR](https://github.com/Cosmo-Tech/webapp-component-ui/pull/313/files) for an example of "default values patching" :slightly_smiling_face: )

csm-thu: Note: another work-around would be to add fallback values with the `??` operator every time we use the `labels` prop, but it seems less readable and harder to maintain than a single dict of all default values

src/components/ScenarioParameters/components/ScenarioParametersInputs/components/TableExportDialog/TableExportDialog.js
csm-thu: Good point, I renamed it to `exportDescription` to be extra unique :smile: (which is convenient when `grep`'ping through the code)

src/components/ScenarioParameters/components/ScenarioParametersInputs/components/TableExportDialog/TableExportDialog.js
esasova: I agree, it will be messy with ?? operators everywhere and patching can be useful when there are lots of labels. +1 to keep props required, in my opinion, it won't be changed a lot (not easy to find an alternative to _Cancel_ :rofl: )

PR #544, opened 2023-02-22
refactor: rename COSMOTECH_URL to ORGANIZATION_URL

csm-thu:


ardvi (approved):
Little comment but LGTM :+1:

src/config/HelpMenuConfiguration.json
```diff
@@ -1,6 +1,6 @@
 {
   "APP_VERSION": null,
-  "COSMOTECH_URL": "https://cosmotech.com",
+  "ORGANIZATION_URL": "https://cosmotech.com",
```
esasova, line 3: We don't need to change it in _HelpMenuConfiguration.vanilla.json_ ? :slightly_smiling_face:

esasova (approved):
One question but LGTM :+1:

src/config/HelpMenuConfiguration.json
csm-thu: You're right, I'll add this too, thank you ! :smile:

PR #546, opened 2023-02-27
Emi/new input

COSMO-EMI:
- [ ] Code is clean


src/components/ScenarioParameters/components/ScenarioParametersInputs/GenericTextInput.js
```diff
@@ -11,13 +11,13 @@ export const GenericTextInput = ({ parameterData, context, parameterValue, setPa
   const { t } = useTranslation();
   const textFieldProps = {
     disabled: !context.editMode,
-    id: parameterData.id,
+    id: `text-input-${parameterData.id}`,
   };
 
   return (
     <BasicTextInput
       key={parameterData.id}
```
csm-thu, line 19: `dataCy` seems to be equal to the `key` prop most of time, maybe we could remove this prop to make their usage easier ?
Also, we should try to determine the best way to declare these `data-cy` values (is this the responsibility of the webapp or the UI components ?) This would be a breaking change, so we should find the best way now :sweat_smile:

PR #556, opened 2023-03-08
feat: [PROD-9525] bold modified parameters

esasova:
- [ ] Code is clean

### DIFF
--- a/src/components/ScenarioParameters/components/ScenarioParametersTabs/ScenarioParameterInput.js
+++ b/src/components/ScenarioParameters/components/ScenarioParametersTabs/ScenarioParameterInput.js
@@ -35,9 +35,9 @@ const ScenarioParameterInput = ({ parameterData, context }) => {
   return (
     <Controller
       name={parameterData.id}
-      render={({ field }) => {
+      render={({ field, fieldState }) => {
         const { value: parameterValue, onChange: setRhfValue } = field;
-
+        const { isDirty } = fieldState;
         const setParameterValue = (newValue) => {
           if (scenarioIdOnMount.current === getCurrentScenarioId()) {
             setRhfValue(newValue);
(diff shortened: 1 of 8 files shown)


csm-thu (commented):
Do you know if it would be possible to move the style definition to the UI components package, and use a common prop (e;g. `dirty`) for all components ? The current solution seems to lead to UI-implementation details being known in the webapp (e.g. the style that contains `'& input'`): it means that changing the UI component implementation would break the style defined in the webapp, whereas it could be a non-breaking change if we use a `dirty` prop instead.

esasova:
Yes, it's possible to apply a class conditionally, good idea, thanks ! Is it better to introduce a new non required prop `isDirty` or we can pass it with all `textFieldProps` ?

csm-thu:
> Yes, it's possible to apply a class conditionally, good idea, thanks ! Is it better to introduce a new non required prop `isDirty` or we can pass it with all `textFieldProps` ?
@esasova a new non-required prop `isDirty` in all generic component would be nice :slightly_smiling_face:

src/components/ScenarioParameters/components/ScenarioParametersTabs/ScenarioParameterInput.js
```diff
@@ -35,9 +35,10 @@ const ScenarioParameterInput = ({ parameterData, context }) => {
   return (
     <Controller
       name={parameterData.id}
-      render={({ field }) => {
+      render={({ field, fieldState }) => {
         const { value: parameterValue, onChange: setRhfValue } = field;
-
+        const { isDirty } = fieldState;
+        const context = { ...tabContext, isDirty };
```
ardvi, line 41: Same remarq than @csm-thu  , i think is better to add isDirty as a optional prop to all generic input, use it in UI for let basic components apply their particular style.

ardvi (commented):
Same remarq as  @csm-thu

csm-thu (approved):
LGTM :+1:

PR #557, opened 2023-03-10
chore: fix possible MSAL error showing on some webapps after enabling 2FA

csm-thu:
`'self'` has been added to the list of allowed `frame-src`


ardvi (approved):
LGTM :+1:

PR #558, opened 2023-03-13
feat: [PROD-9527] implement warning if powerBI dashboard isn't sync

esasova:
- [ ] Code is clean

### DIFF
--- a/cypress/commons/actions/generic/ScenarioSelector.js
+++ b/cypress/commons/actions/generic/ScenarioSelector.js
@@ -54,10 +54,13 @@ function checkValidationStatusInScenarioSelector(searchStr, scenarioId, expected
 }
 
 function writeInScenarioSelectorInput(searchStr) {
-  return getScenarioSelector()
+  return getScenarioSelectorInput()
     .click()
     .should('not.be.disabled')
-    .type('{selectAll}{backspace}' + searchStr); // clear() does not always work, use "{selectAll}{backspace}" instead
+    .type(
+      '{selectAll}{backspace}' + searchStr, // clear() does not always work, use "{selectAll}{backspace}" instead
+      { force: true } // Force click to handle cases when the error banner is displayed above the selector options
+    );
 }
 
 // Parameters:


src/views/Scenario/Scenario.js
```diff
@@ -42,7 +56,10 @@ const Scenario = () => {
   const { t } = useTranslation();
 
   // RHF
-  const methods = useForm();
+  const {
+    formState: { isDirty },
+    ...methods
+  } = useForm();
```
csm-thu, line 62: Doesn't this change remove `formState` from the `methods` object ? Isn't it an issue when we use `<FormProvider {...methods}>` below?

src/views/Scenario/Scenario.js
```diff
@@ -228,6 +245,30 @@ const Scenario = () => {
...
     validationStatus: scenarioValidationStatusLabels,
   };
+  // FOR DEVELOPMENT PURPOSE ONLY, DELETE BEFORE MERGE
+  const lastRun = '2023-03-13T16:25:30.795486885Z';
+  // FOR DEVELOPMENT PURPOSE ONLY, DELETE BEFORE MERGE
+
+  const hasScenarioBeenRun = useMemo(
+    () => currentScenario?.data?.lastRun !== null && currentScenario?.data?.state === 'Successful',
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    [currentScenario?.data?.lastRun]
```
csm-thu, line 255: `currentScenario?.data?.state` is missing from the dependency array, was it done on purpose?

src/views/Scenario/Scenario.js
```diff
@@ -228,6 +245,30 @@ const Scenario = () => {
...
+  );
+  const isDashboardSync = useMemo(
+    () => new Date(currentScenario?.data?.lastUpdate) < new Date(lastRun),
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    [currentScenario?.data?.lastUpdate]
```
csm-thu, line 260: `lastRun` is missing from the dependency array, was it done on purpose? (even though lastRun is currently hard-coded, it won't be hard-coded when the missing calls to the API will be added: we will probably need it in the dependency array then)

src/views/Scenario/Scenario.js
```diff
@@ -228,6 +245,30 @@ const Scenario = () => {
...
+  );
+  const getDashboardCardStyle = () => {
+    return hasScenarioBeenRun
+      ? !isDirty && isDashboardSync
+        ? {
+            px: 3,
+            '&:last-child': { pb: 2 },
+          }
+        : { px: 3, bgcolor: 'dashboard.warning', '&:last-child': { pb: 2 } }
```
csm-thu, line 269: Both objects seem to be very close except for the `bgcolor` key, can we factorize the common style? (maybe we could create both constants above the component definition?)

src/views/Scenario/Scenario.js
esasova: I decided to use only one dependency because `currentScenario?.data?.state` is closely linked to `currentScenario?.data?.lastRun` (it changes only after a run) so no need to make a double check but if it can be necessary, I will put it in dependency array.

src/views/Scenario/Scenario.js
esasova: The same, I thought it wasn't necessary because an update is always performed before a run but I'll put it in the array :+1:

ardvi (commented):
Some remarqs

src/views/Scenario/Scenario.js
```diff
@@ -286,8 +323,19 @@ const Scenario = () => {
             </Card>
           </Grid>
           <Grid item xs={12}>
-            <Card component={Paper}>
-              <CardContent>
+            <Card component={Paper} sx={{ p: 0 }}>
```
ardvi, line 326: I think in fact, it's better to move the card management in ScenarioPowerBiReport component instead of manage this in Scenario view.

src/views/Scenario/Scenario.js
csm-thu: Currently, we have no need to try to optimize performance early, we can probably add both in the dependencies array :slightly_smiling_face: 
Furthermore, adding only one of them seems like a risk for bugs when we change the code in the future :smile:

src/views/Scenario/Scenario.js
```diff
@@ -228,7 +243,29 @@ const Scenario = () => {
     label: scenarioListLabel,
     validationStatus: scenarioValidationStatusLabels,
   };
+  // FOR DEVELOPMENT PURPOSE ONLY, DELETE BEFORE MERGE
+  const lastRun = '2023-03-13T16:25:30.795486885Z';
+  // FOR DEVELOPMENT PURPOSE ONLY, DELETE BEFORE MERGE
 
+  const hasScenarioBeenRun = useMemo(
+    () => currentScenario?.data?.lastRun !== null && currentScenario?.data?.state === 'Successful',
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
csm-thu, line 252: Do we still need the exception `eslint-disable-next-line`? Idem for the next `useMemo` below

src/views/Scenario/Scenario.js
```diff
@@ -228,7 +243,29 @@ const Scenario = () => {
...
+    [currentScenario?.data?.lastUpdate, lastRun]
+  );
+  const dashboardCardStyle = hasScenarioBeenRun
+    ? {
+        px: 3,
+        '&:last-child': { pb: 2 },
+        bgcolor: () => {
+          if (isDirty || !isDashboardSync) return 'dashboard.warning';
+        },
```
csm-thu, line 266: Is it ok to have a function for `bgcolor`? In your previous version, it wasn't a function but a string: `{ px: 3, bgcolor: 'dashboard.warning', '&:last-child': { pb: 2 } }`

esasova: I think it's ok because the function returns a string, can it be a problem ?

csm-thu: I just checked in mui docs, and it seems it can work indeed :+1: 
Yet, we could probably make the code simpler by replacing the function by:
`bgcolor: (isDirty || !isDashboardSync) ? 'dashboard.warning' : undefined;`

nborde-CSM (approved):
LGTM :+1

csm-thu (approved):
LGTM :+1: 
(but do not merge until we remove the hard-coded value :smile:)

PR #559, opened 2023-03-15
Thu/fix instance view always visible

csm-thu:


cypress/integration/brewery/InstanceView.spec.js
```diff
@@ -2,14 +2,40 @@
...
 
 Cypress.Keyboard.defaults({
   keystrokeDelay: 0,
 });
 
-describe('Instance view', () => {
+const BASE_URL = Cypress.config().baseUrl;
+const url = BASE_URL + '/' + DEFAULT_WORKSPACE.id;
+
+describe('Instance view disabled', () => {
```
ardvi, line 18: Sure to add another cypress test instead of a jest of getTabsForCurrentWorkspace ?

ardvi (approved):
Just a comment but LGTM :+1:

cypress/integration/brewery/InstanceView.spec.js
csm-thu: In my opinion, testing the `getTabsForCurrentWorkspace` function would only test a filter on an array (and maybe the `ConfigUtils.isInstanceViewConfigValid` function if it is not already tested elsewhere).
The cypress test, on the other hand, may catch many more bugs:
- reception of the workspace data by the webapp (especially the `instanceView` object, when `undefined`)
- merging workspace data with the "overrides" config
- make sure that changes when the filter is applied correctly leads to the tab not being shown

Furthermore, the new cypress test can be fully stubbed and does not run simulations, it shouldn't add much time to the existing tests :slightly_smiling_face:

nborde-CSM (approved):
LGTM :+1:

PR #565, opened 2023-03-28
Nbo/save discard and launch prod 7902

nborde-CSM:
- [ ] Code is clean


public/locales/en/translation.json
```diff
@@ -38,11 +38,12 @@
...
             "noTabs": "No parameters to edit"
           },
-          "discard": "Discard modifications",
+          "discard": "DISCARD",
           "edit": "Edit parameters",
-          "launch": "Launch scenario",
+          "launch": "LAUNCH",
           "update": {
-            "launch": "Update and launch scenario"
+            "update": "SAVE",
```
csm-thu, line 45: `scenario.parameters.update.update` is redundant, maybe we could move this label up to `scenario.parameters.save` ? This way, it would be at the same level as the other actions labels (discard, launch)

public/locales/en/translation.json
```diff
@@ -193,7 +194,6 @@
             "label": "Run status:",
             "successful": "Successful",
             "running": "Running",
-            "dataingestioninprogress": "Transferring results",
```
csm-thu, line 196: Why did  you remove this label ? Maybe it was a failed merge or rebase, could you check it please ?

src/components/ScenarioParameters/ScenarioParameters.js
```diff
@@ -156,14 +164,19 @@ const ScenarioParameters = ({
...
   useEffect(() => {
     discardLocalChanges();
     // eslint-disable-next-line
-  }, [currentScenarioData?.id]);
+  }, [scenarioId]);
+
+  useEffect(() => {
+    if (scenarioStatus === STATUSES.SUCCESS) reset({ ...getValues() });
+    // eslint-disable-next-line
+  }, [scenarioStatus]);
```
csm-thu, line 179: `reset({ ...getValues() });` is already called when the scenario is launched, do we really need it here too?

src/components/ScenarioParameters/ScenarioParameters.js
```diff
@@ -176,100 +189,71 @@ const ScenarioParameters = ({
...
-
-  const [showWarningBeforeLaunchPopup, setShowWarningBeforeLaunchPopup] = useState(false);
-  const closeConfirmLaunchDialog = () => {
-    setShowWarningBeforeLaunchPopup(false);
-  };
-
   const updateBeforeLaunch = useRef(null);
 
-  const confirmAndLaunch = (event, updateBeforeLaunch_) => {
+  const scenarioLaunch = (event, updateBeforeLaunch_) => {
```
csm-thu, line 194: Functions names should start with a verb, could we rename it? (we already have `startScenarioLaunch` though, I don't know what a good name would be here).
_Idem_ for function `scenarioSave` below.

csm-thu (changes requested):

public/locales/en/translation.json
```diff
@@ -38,11 +38,12 @@
...
           },
-          "discard": "Discard modifications",
+          "discard": "DISCARD",
           "edit": "Edit parameters",
-          "launch": "Launch scenario",
+          "launch": "LAUNCH",
           "update": {
-            "launch": "Update and launch scenario"
+            "update": "SAVE",
+            "launch": "SAVE AND LAUNCH"
```
esasova, line 46: Do we still have the button update and launch ? If it's necessary for older versions, why change the name ?

cypress/commons/actions/generic/ScenarioParameters.js
```diff
@@ -39,27 +36,10 @@ function getParametersConfirmDiscardButton() {
...
-function getLaunchCancelButton() {
-  return cy.get(GENERIC_SELECTORS.scenario.parameters.dialogLaunch.cancelButton);
-}
-function getDontAskAgainCheckbox() {
-  return cy.get(GENERIC_SELECTORS.scenario.parameters.dialogLaunch.dontAskAgainCheckbox).find('input');
-}
-function checkDontAskAgain() {
-  getDontAskAgainCheckbox().check();
+function getSaveButton(timeout) {
+  return cy.get(GENERIC_SELECTORS.scenario.parameters.saveButton, timeout);
```
csm-thu, line 40: Watch the type of the `timeout` parameter. To be as simple as possible, and to be consistent with other functions, we probably want it to be a **number of seconds**, so you need to use `timeout ? { timeout: timeout * 1000 }` (or just `{ timeout: timeout * 1000 }`) instead of only `timeout` when calling `cy.get`.

Also, if we never need to have a specific timeout value when getting the save button, you can remove the `timeout` parameter

src/components/ScenarioParameters/ScenarioParameters.js
```diff
@@ -176,100 +189,72 @@ const ScenarioParameters = ({
...
-    );
-  } else if (noTabsShown) {
-    disabledEditTooltip = t(
-      'commoncomponents.button.scenario.parameters.disabledEditTooltip.noTabs',
-      'No parameters to edit'
-    );
-  }
+  // console.log(currentScenarioStatus);
+  // console.log(isCurrentScenarioSaving);
+
```
esasova, line 249: remaining logs to remove :slightly_smiling_face:

src/components/ScenarioParameters/components/SaveLaunchDiscardButton.js
```diff
@@ -0,0 +1,77 @@
...
+import { Grid, Button, Typography } from '@mui/material';
+import { PermissionsGate } from '@cosmotech/ui';
+import SaveIcon from '@mui/icons-material/Save';
+import CancelIcon from '@mui/icons-material/Cancel';
+import PlayCircleOutlineIcon from '@mui/icons-material/PlayCircleOutline';
+import { useTranslation } from 'react-i18next';
+import { ACL_PERMISSIONS } from '../../../services/config/accessControl';
+import { useUserAppAndCurrentScenarioPermissions } from '../../../hooks/SecurityHooks';
+
+const SaveLaunchDiscardButton = ({ save, launch, saveAndLauch, discard, isDirty, runDisabled }) => {
```
esasova, line 15: question about naming: usually we begin function's name in prop by _on_ (`onSave`, `onLaunch`, `onDiscard` etc) and boolean props begin with _is_ or _has_, isn't it relevant here as well ?

src/components/ScenarioParameters/ScenarioParameters.js
```diff
@@ -176,100 +189,71 @@ const ScenarioParameters = ({
...
     updateBeforeLaunch.current = updateBeforeLaunch_;
-    if (localStorage.getItem('dontAskAgainToConfirmLaunch') === 'true') {
-      startScenarioLaunch();
-    } else {
-      setShowWarningBeforeLaunchPopup(true);
-    }
+    startScenarioLaunch();
+  };
+
+  const scenarioSave = async (event) => {
```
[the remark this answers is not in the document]
nborde-CSM: I added await processFilesParameters();  which is not yet visible.

src/components/ScenarioParameters/components/SaveLaunchDiscardButton.js
```diff
@@ -0,0 +1,77 @@
...
+              <Button
+                data-cy="launch-scenario-button"
+                variant="contained"
+                disabled={runDisabled}
+                startIcon={<PlayCircleOutlineIcon />}
+                onClick={isDirty ? saveAndLauch : launch}
+              >
+                <Typography variant="inherit">
+                  {isDirty
+                    ? t('commoncomponents.button.scenario.parameters.update.launch', 'SAVE & LAUNCH ')
```
esasova, line 56: I can't find _SAVE AND LAUNCH_ button in mockups, I thought it was just _SAVE_ for this feature :thinking:

src/components/ScenarioParameters/ScenarioParameters.js
nborde-CSM: This reset is used to correctly handle parameters reset after save.
In fact, it is the reset during scenario launch which is now redundant.

src/components/ScenarioParameters/ScenarioParameters.js
```diff
@@ -50,32 +51,34 @@ const getRunTemplateParametersIds = (runTemplatesParametersIdsDict, runTemplateI
...
-  accordionSummaryExpanded,
-}) => {
+const ScenarioParameters = ({ onChangeAccordionSummaryExpanded, accordionSummaryExpanded }) => {
   const classes = useStyles();
   const { t } = useTranslation();
   const {
     scenariosData,
     datasetsData,
     addDatasetToStore,
+    currentScenario,
```
esasova, line 61: We use `currentScenario` prop only to get its status, maybe it's better to create a `currentScenarioStatus` hook to avoid getting whole current scenario data twice (in `currentScenario` and `currentScenarioData`) ?

src/components/ScenarioParameters/ScenarioParameters.js
```diff
@@ -176,100 +184,71 @@ const ScenarioParameters = ({
...
-      'This scenario is validated, you cannot edit its parameters'
-    );
-  } else if (noTabsShown) {
-    disabledEditTooltip = t(
-      'commoncomponents.button.scenario.parameters.disabledEditTooltip.noTabs',
-      'No parameters to edit'
-    );
-  }
+  // console.log(currentScenarioStatus);
+  // console.log(isCurrentScenarioSaving);
```
csm-thu, line 248: Commented code to remove

src/components/ScenarioParameters/ScenarioParameters.js
csm-thu: If we remove the whole block, then `scenarioStatus` and `currentScenario` can be removed from the hook as well, I think?

src/components/ScenarioParameters/components/SaveLaunchDiscardButton.js
nborde-CSM: I have talked about this Emma and she validated this initiative.
So, when parameters form is dirty, the button display SAVE AND LAUNC.

src/components/ScenarioParameters/ScenarioParameters.js
nborde-CSM: I think it is more simple to keep only currentScenario prop and make several const _currentScenarioData_ and _currentScenarioStatus_ from this prop.

src/components/ScenarioParameters/components/ScenarioParametersInputs/GenericUploadFile.js
```diff
@@ -68,7 +68,7 @@ export const GenericUploadFile = ({
         FileManagementUtils.downloadFile(organizationId, workspaceId, datasetId, setClientFileDescriptorStatus);
       }}
       file={parameter}
-      editMode={context.editMode}
+      editMode={!context.editMode}
```
esasova, line 71: `editMode` doesn't exist anymore so `GenericUploadFile` is always in `editMode` even when scenario is running

src/components/ScenarioParameters/components/ScenarioParametersInputs/GenericTable.js
```diff
@@ -492,7 +492,7 @@ export const GenericTable = ({
         labels={tableLabels}
         tooltipText={t(TranslationUtils.getParameterTooltipTranslationKey(parameterData.id), '')}
         dateFormat={dateFormat}
-        editMode={context.editMode}
+        editMode={!!context.editMode}
```
esasova, line 495: probably a typo: !! (or you are using a conversion method stackoverflow didn't recommended already 13 years ago :rofl:)

cypress/integration/brewery/WorkingInTwoWorkspaces.spec.js
```diff
@@ -235,44 +231,44 @@ describe('Switching between workspaces and running four scenarios at the same ti
...
+            BreweryParameters.getStockInput().should('value', '400');
+            BreweryParameters.getRestockInput().should('value', '10');
+            BreweryParameters.getWaitersInput().should('value', '15');
 
             // browse to the second scenario and check its parameters
             route.browse({ url: `${BREWERY_WORKSPACE_ID}/scenario/${secondWorkspaceSharedScenarioId}` });
-            BreweryParameters.getStock().should('have.text', '400');
-            BreweryParameters.getRestock().should('have.text', '10');
-            BreweryParameters.getWaiters().should('have.text', '15');
+            BreweryParameters.getStockInput().should('value', '400');
```
esasova, line 269: The same, we may consider adding a timeout here: test breaks because cypress tries to check the value before changing the good scenario

src/components/ShareCurrentScenarioButton/ShareCurrentScenarioButton.js
```diff
@@ -2,12 +2,13 @@
...
 
 import React from 'react';
+import PropTypes from 'prop-types';
 import { PermissionsGate, RolesEditionButton } from '@cosmotech/ui';
 
 import { ACL_PERMISSIONS } from '../../services/config/accessControl';
 import { useShareCurrentScenarioButton } from './ShareCurrentScenarioButtonHook';
 
-const ShareCurrentScenarioButton = () => {
+const ShareCurrentScenarioButton = ({ disabled }) => {
```
esasova, line 11: IIRC we can avoid adding this prop because the component can access isDirty state directly bu `useFormState() `hook

csm-thu: Done :slightly_smiling_face:

cypress/commons/actions/generic/ScenarioParameters.js
```diff
@@ -39,27 +37,10 @@ function getParametersConfirmDiscardButton() {
...
-function getLaunchCancelButton() {
-  return cy.get(GENERIC_SELECTORS.scenario.parameters.dialogLaunch.cancelButton);
-}
-function getDontAskAgainCheckbox() {
-  return cy.get(GENERIC_SELECTORS.scenario.parameters.dialogLaunch.dontAskAgainCheckbox).find('input');
-}
-function checkDontAskAgain() {
-  getDontAskAgainCheckbox().check();
+function getSaveButton(timeout) {
+  return cy.get(GENERIC_SELECTORS.scenario.parameters.saveButton, timeout ? { timeout: timeout * 1000 } : undefined);
```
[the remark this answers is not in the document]
csm-thu: A timeout set to 0 means that the element must be immediately `get`-able, without any delay for network requests or for react to render the changes. Instead of 0, I usually set the default value to 4 seconds, the [default timeout defined by cypress](https://docs.cypress.io/guides/references/configuration#Timeouts) :slightly_smiling_face: 

Using the ternary instead of setting the timeout to 4 decreases readability (IMO), but it has the advantage of not defining a specific timeout here, and thus fall back to cypress configuration (if we want to change the default timeout someday, we'll only have to change the cypress configuration file instead of editing the default values set in all actions using a timeout)

Until we discuss this subject and decide what should be the "best practice", we can accept both :smile:

csm-thu (approved):
I committed some improvements in additional commits, branch could probably be merged now (after fixup of my commits)

esasova (approved):

ardvi (approved):
LGTM :+1:

PR #576, opened 2023-04-19
test: add test to check forced update of scenario metadata parameters

csm-thu:


cypress/integration/brewery/ScenarioMetadataParameters.spec.js
```diff
@@ -0,0 +1,52 @@
+import { Login, ScenarioParameters } from '../../commons/actions';
```
Guismus, line 1: Missing header at the start of file

csm-thu: Done :+1:

nborde-CSM (approved):
One remark but LGTM :+1:

cypress/integration/brewery/ScenarioMetadataParameters.spec.js
```diff
@@ -0,0 +1,55 @@
...
+import {
+  SCENARIOS,
+  SOLUTIONS,
+  SCENARIO_METADATA_PARAMETERS_IDS,
+} from '../../fixtures/stubbing/ScenarioMetadataParameters';
+import { stub } from '../../commons/services/stubbing';
+import { apiUtils } from '../../commons/utils';
+
+const validateScenarioUpdateRequest = (req) => {
```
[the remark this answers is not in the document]
csm-thu: The mechanism to apply a validation function to the interception is already generic and implemented in the apiUtils file (the 2nd parameter of `interceptUpdateScenario`).
The function `validateScenarioUpdateRequest`, on the other hand, aims to check that the intercepted request contains the scenario metadata parameters: it is really specific to the current test and will probably never be necessary for another test.

esasova (approved):
LGTM :+1:

PR #586, opened 2023-04-27
fix: fix console warning regarding undefined required prop

csm-thu:
During transition between scenarios with different run templates, the default value for a given scenario parameter may temporarily not be available. This leads to console warnings because the defaultParameterValue prop was decalred as required. Yet, it does not need to be required: the component could use null or undefined as a default value, or wait for the actual default value to be retrieved during transitions between scenarios.


ardvi (approved):
LGTM :+1:

esasova (approved):
LGTM :+1:

nborde-CSM (approved):
LGTM :+1:

PR #587, opened 2023-04-27
fix: adapt tooltip message on Share button when there are unsaved changes

csm-thu:


public/locales/en/translation.json
```diff
@@ -30,6 +30,9 @@
           }
         }
       },
+      "shareScenario": {
```
ardvi, line 33: it's weird to create a new entry here, all other labels are in commoncomponents.dialog.share.button

src/components/ShareCurrentScenarioButton/ShareCurrentScenarioButton.js
```diff
@@ -22,6 +24,11 @@ const ShareCurrentScenarioButton = () => {
   } = useShareCurrentScenarioButton();
 
   const { isDirty } = useFormState();
+  if (isDirty && shareScenarioDialogLabels?.button?.tooltip)
```
ardvi, line 27: i think it's better to modify getShareScenarioDialogLabels for get the good label

ardvi (commented):
Some comments

src/components/ShareCurrentScenarioButton/ShareCurrentScenarioButton.js
esasova: Maybe, it's better to change it in ui package and add a disabled tooltip label there ? :slightly_smiling_face:  (or make like in CreateScenarioLabels in the function itself, yes)

nborde-CSM (approved):
LGTM :+1:

public/locales/en/translation.json
csm-thu: Done :slightly_smiling_face:

src/components/ShareCurrentScenarioButton/ShareCurrentScenarioButton.js
csm-thu: Done, thank you for the feedback :slightly_smiling_face: 
I moved the call to the `useFormState` hook in the `ShareCurrentScenarioButtonHook` to get the value of `isDirty`, and added it as a parameter of the `getShareScenarioDialogLabels` function.

Modifying the UI component would mean to add a new prop for the form `isDirty` state to choose between possible tooltips labels: it's probably best to keep this logic in the webapp repo and provide only the tooltip message to the generic component.

ardvi (approved):
LGTM :+1:

esasova (approved):
LGTM :+1:

PR #611, opened 2023-06-20
test: fix cypress test TableParameters-files_part3

csm-thu:


esasova (approved):
The test is still brittle but it's better that completely broken so LGTM :+1:

PR #619, opened 2023-06-28
Esas/add badges to tab with a non validated field prod 11781

esasova:
- [ ] Code is clean

### DIFF
--- a/src/components/ScenarioParameters/components/ScenarioParametersTabs/ScenarioParametersTabs.js
+++ b/src/components/ScenarioParameters/components/ScenarioParametersTabs/ScenarioParametersTabs.js
@@ -2,13 +2,14 @@
 // Licensed under the MIT license.
 
 import React, { useEffect, useState } from 'react';
-import { Tab } from '@mui/material';
+import { Badge, Tab } from '@mui/material';
 import makeStyles from '@mui/styles/makeStyles';
 import { TabContext, TabList, TabPanel } from '@mui/lab';
 import { useTranslation } from 'react-i18next';
 import LockIcon from '@mui/icons-material/Lock';
-import { ConfigUtils, TranslationUtils } from '../../../../utils';
+import { ConfigUtils, ScenarioParametersUtils, TranslationUtils } from '../../../../utils';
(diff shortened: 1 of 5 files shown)


src/components/ScenarioParameters/components/ScenarioParametersTabs/ScenarioParametersTabs.js
```diff
@@ -20,19 +21,30 @@ const useStyles = makeStyles((theme) => ({
...
 
-function _buildScenarioTabList(tabs, userRoles, t) {
+function _buildScenarioTabList(tabs, userRoles, t, errors) {
   const tabListComponent = [];
+  const errorsByTab = Object.assign(
+    {},
+    ...tabs.map((tab) => ({
+      [tab.id]: tab.parameters.map((param) => param.id).filter((parameter) => Object.keys(errors).includes(parameter)),
+    }))
+  );
```
csm-thu, line 31: Can we immediately store the number of errors for each tab in this loop (instead of accessing `.length` below), or 0 if there are no errors ?
Maybe the construction of `errorsByTab` could be moved to a distinct function too, its code seems to be complex enough that we might want to write unit tests on it :slightly_smiling_face:

nborde-CSM, line 31: Why don't directly do:
`const errorsByTab = {
    ...tabs.map((tab) => ({
      [tab.id]: tab.parameters.map((param) => param.id).filter((parameter) => Object.keys(errors).includes(parameter)),
    }))
  )};`

nborde-CSM (approved):
One comment but LGTM :+1:

src/components/ScenarioParameters/components/ScenarioParametersTabs/ScenarioParametersTabs.js
esasova: I tried but it doesn't spread the array correctly :slightly_frowning_face:

csm-thu (approved):
LGTM :+1:

PR #628, opened 2023-07-12
feat: [PROD-11662] POC add constraint to another paramter for dates

esasova:
- [ ] Code is clean

### DIFF
--- a/cypress/fixtures/stubbing/ScenarioParametersValidation/index.js
+++ b/cypress/fixtures/stubbing/ScenarioParametersValidation/index.js
@@ -2,4 +2,4 @@
 // Licensed under the MIT license.
 
 export { SCENARIOS } from './scenarios';
-export { SOLUTIONS } from './solutions';
+export { SOLUTIONS, SOLUTIONS_WITH_WRONG_CONSTRAINT } from './solutions';
--- a/src/hooks/ParameterConstraintsHooks.js
+++ b/src/hooks/ParameterConstraintsHooks.js
@@ -1,0 +1,103 @@
+// Copyright (c) Cosmo Tech.
+// Licensed under the MIT license.
+
+import { useSolutionParameters } from '../state/hooks/SolutionHooks';
+import { TranslationUtils, ParameterConstraintsUtils } from '../utils';
+import isBefore from 'date-fns/isBefore';
+import isAfter from 'date-fns/isAfter';
+import isSameDay from 'date-fns/isSameDay';
+import { useCallback, useEffect } from 'react';
+import { useFormContext } from 'react-hook-form';
+import { useTranslation } from 'react-i18next';
+
+export const useParameterConstraint = (parameterData) => {
+  const { t } = useTranslation();
+  const parametersList = useSolutionParameters();
+  const { trigger, watch } = useFormContext();
+  const constraint = ParameterConstraintsUtils.getParameterValidationConstraint(
+    parameterData?.options?.validation,
+    parameterData.varType,
+    parametersList
+  );
+  const constrainingValue = constraint ? watch(constraint?.id) : undefined;
+
+  useEffect(() => {
+    if (constraint) {
+      trigger(parameterData.id);
+    }
+  }, [constraint, parameterData.id, trigger, constrainingValue]);
+  return { t, constraint, constrainingValue };
+};
+
+export const useDateConstraintValidation = (parameterData) => {
+  const { t, constraint, constrainingValue } = useParameterConstraint(parameterData);
+
+  const getDateConstraintValidation = useCallback(
+    (value) => {
+      const getTranslationLabel = (translationString, fallbackString) => {
+        return t(translationString, fallbackString, {
+          field: t(TranslationUtils.getParameterTranslationKey(parameterData?.id)),
+          constraint: t(TranslationUtils.getParameterTranslationKey(constraint?.id)),
+        });
+      };
+
+      if (constraint) {
+        switch (constraint.operator) {
+          case '>':
+            return (
+              isAfter(new Date(value), new Date(constrainingValue)) ||
+              getTranslationLabel(
+                'views.scenario.scenarioParametersValidationErrors.after',
+                'The field {{field}} must be strictly after the field {{constraint}}'
+              )
+            );
+          case '<':
+            return (
+              isBefore(new Date(value), new Date(constrainingValue)) ||
+              getTranslationLabel(
+                'views.scenario.scenarioParametersValidationErrors.before',
+                'The field {{field}} must be strictly before the field {{constraint}}'
+              )
+            );
+          case '>=':
+            return (
+              !isBefore(new Date(value), new Date(constrainingValue)) ||
+              getTranslationLabel(
+                'views.scenario.scenarioParametersValidationErrors.afterOrEqual',
+                'The field {{field}} must be after or equal to the field {{constraint}}'
+              )
+            );
+          case '<=':
+            return (
+              !isAfter(new Date(value), new Date(constrainingValue)) ||
+              getTranslationLabel(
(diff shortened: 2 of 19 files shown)


src/utils/SolutionsUtils.js
```diff
@@ -55,9 +56,26 @@ const patchSolutionIfLocalConfigExists = async (originalSolution) => {
...
 
+const checkParametersValidationConstraintsInSolution = (data) => {
+  const parametersWithConstraint = data?.parameters?.filter((parameter) => parameter.options?.validation);
+  parametersWithConstraint?.forEach((parameter) => {
+    const constraint = ParameterConstraintsUtils.getParameterValidationConstraint(
+      parameter.options.validation,
+      parameter.varType,
+      data.parameters
+    );
+    if (constraint === null)
```
csm-thu, line 67: When the constraint configuration is incorrect, maybe we should also delete the constraint from the config and "patch" the solution that will be used in the webapp? This could prevent some bugs with RHF trying to read invalid constraints

src/utils/ParameterConstraintsUtils.js
```diff
@@ -0,0 +1,34 @@
...
+};
+
+const getVarTypesComparison = (varType, varTypeToCompare) => {
+  const numberVarTypes = ['int', 'number'];
+  return varType === varTypeToCompare || [varType, varTypeToCompare].every((el) => numberVarTypes.includes(el));
+};
+
+const getParameterValidationConstraint = (validationString, varType, parametersList) => {
+  if (validationString) {
+    const [operator, id] = validationString.replace(/\s+/g, ' ').trim().split(' ');
```
csm-thu, line 19: Maybe we should return `null` early if `id` is `undefined` at this point? (when the constraint string has no "space" character inside it)

I fear that proceeding to the lines below with an `undefined` id may lead to edge cases and bugs

src/hooks/ParameterConstraintsHooks.js
```diff
@@ -0,0 +1,109 @@
...
+    parametersList
+  );
+  const constrainingValue = constraint ? watch(constraint?.id) : undefined;
+
+  useEffect(() => {
+    if (constraint) {
+      trigger(parameterData.id);
+    }
+  }, [constraint, parameterData.id, trigger, constrainingValue]);
+  return { t, constraint, constrainingValue };
```
csm-thu, line 31: It seems that the `t` function is not used in `useParameterConstraint`, do we need to retrieve it and return it here? (or maybe you were anticipating future hooks, to factorize the code between several types of constraints)

csm-thu (approved):
Suggested minor changes but LGTM :+1:

cypress/fixtures/stubbing/ScenarioParametersValidation/index.js
```diff
@@ -3,3 +3,4 @@
 
 export { SCENARIOS } from './scenarios';
 export { SOLUTIONS } from './solutions';
+export { SOLUTIONS_WITH_WRONG_CONSTRAINT } from './solutions';
```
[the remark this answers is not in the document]
esasova: Nice catch, thanks ! :slightly_smiling_face:

src/utils/__test__/SolutionsUtils.spec.js
```diff
@@ -94,4 +94,50 @@ describe('addRunTemplatesParametersIdsDict for a minimal or incomplete solution'
...
+        {
+          id: 'parameter2',
+          varType: 'string',
+          options: null,
+        },
+      ],
+    };
+    SolutionsUtils.checkParametersValidationConstraintsInSolution(solution);
+    expect(spyConsoleWarn).toHaveBeenCalledTimes(1);
+    expect(spyConsoleWarn).toHaveBeenCalledWith(warning);
```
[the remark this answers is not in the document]
esasova: Great idea, it wasn't easy to hard code **exactly** the same message and I was afraid it would be brittle, thanks for this more robust suggestion :slightly_smiling_face:

src/hooks/ParameterConstraintsHooks.js
esasova: Yes, I anticipated it for future hooks (numbers, string etc) because I wanted to avoid declaring `const { t } = useTranslation() `in every hook and since all hooks will use `useParameterConstraint`, I included `t` in its returned value :slightly_smiling_face:

src/utils/ParameterConstraintsUtils.js
```diff
@@ -0,0 +1,35 @@
...
+  bool: ['==', '!='],
+};
+
+const getIsVarTypesComparisonValid = (varType, varTypeToCompare) => {
+  const numberVarTypes = ['int', 'number'];
+  return varType === varTypeToCompare || [varType, varTypeToCompare].every((el) => numberVarTypes.includes(el));
+};
+
+const getParameterValidationConstraint = (validationString, varType, parametersList) => {
+  if (validationString) {
```
nborde-CSM, line 18: Do you realy want to test if _validationString_ is empty ?
In the case where you only want to test if _validationString_ is _null_ or _undefined_, preferably use _!= null_ test.

src/utils/ParameterConstraintsUtils.js
```diff
@@ -0,0 +1,35 @@
...
+    const [operator, id] = validationString.replace(/\s+/g, ' ').trim().split(' ');
+    if (id === undefined) return null;
```
nborde-CSM, line 20: What about _id_ is equal to _null_ ?

src/utils/ParameterConstraintsUtils.js
```diff
@@ -0,0 +1,35 @@
...
+    const varTypeToCompare = Array.isArray(parametersList)
+      ? parametersList?.find((parameter) => parameter.id === id)?.varType
+      : null;
+    const isVarTypesComparisonValid = getIsVarTypesComparisonValid(varType, varTypeToCompare);
+    if (operatorsDict[varType]?.includes(operator) && isVarTypesComparisonValid) {
```
nborde-CSM, line 25: It seems you do not reuse the const _isVarTypesComparisonValid_ in the function. So, should you not use directly _getIsVarTypesComparisonValid_ function here?

src/utils/SolutionsUtils.js
```diff
@@ -55,9 +56,28 @@ const patchSolutionIfLocalConfigExists = async (originalSolution) => {
...
 
+const checkParametersValidationConstraintsInSolution = (data) => {
+  const parametersWithConstraint = data?.parameters?.filter((parameter) => parameter.options?.validation);
+  parametersWithConstraint?.forEach((parameter) => {
+    const constraint = ParameterConstraintsUtils.getParameterValidationConstraint(
+      parameter.options.validation,
+      parameter.varType,
+      data.parameters
+    );
+    if (constraint === null) {
```
nborde-CSM, line 67: What about _constraint_ is about to _undefined_

nborde-CSM (approved):
Some questions but LGTM :+1:

src/utils/ParameterConstraintsUtils.js
esasova: I prefer test it in case there is an arror in solution configuration (something like: `validation: ''`) :slightly_smiling_face:

src/utils/ParameterConstraintsUtils.js
esasova: `split` function cannot return `null`, it will return `undefined` (if there is no spaces in validation string for example: `validation: '>parameter'`) otherwise it will be a string

src/utils/ParameterConstraintsUtils.js
esasova: I create this variable for better readability but we can use the function directly indeed :slightly_smiling_face:

PR #653, opened 2023-08-30
Nbo/stop scenario running before delete prod 12104

nborde-CSM:
- [ ] Code is clean

### DIFF
--- a/cypress/commons/utils/apiUtils.js
+++ b/cypress/commons/utils/apiUtils.js
@@ -149,6 +149,24 @@ const interceptLaunchScenario = (stubbingOptions) => {
   return alias;
 };
 
+const interceptStopScenarioRun = (scenarioId) => {
+  const alias = forgeAlias('reqStopScenarioRun');
+  cy.intercept({ method: 'POST', url: API_REGEX.STOP_SCENARIO_RUN, times: 1 }, (req) => {
+    if (stub.isEnabledFor('LAUNCH_SCENARIO')) {
+      const scenarioRunId = req.url.match(API_REGEX.SCENARIO_RUN)[1];
+      const scenarioLastRun = {
+        id: scenarioRunId,
+        endTime: null,
+        message: null,
+        estimatedDuration: null,
+      };
+
+      req.reply(scenarioLastRun);
+    }
+  }).as(alias);
+  return alias;
+};
+
 const interceptGetScenarioRun = () => {
   const alias = forgeAlias('reqGetScenarioRun');
   cy.intercept({ method: 'GET', url: API_REGEX.SCENARIO_RUN, times: 1 }, (req) => {
@@ -341,6 +359,7 @@ export const apiUtils = {
   interceptAuthentication,
   interceptCreateScenario,
   interceptLaunchScenario,
+  interceptStopScenarioRun,
   interceptGetScenarioRun,
   interceptGetScenarioRunStatus,
   interceptUpdateScenario,
--- a/cypress/e2e/brewery/StopScenarioRun.cy.js
+++ b/cypress/e2e/brewery/StopScenarioRun.cy.js
@@ -1,29 +1,41 @@
 // Copyright (c) Cosmo Tech.
 // Licensed under the MIT license.
 
-import { Login } from '../../commons/actions/brewery';
+import { stub } from '../../commons/services/stubbing';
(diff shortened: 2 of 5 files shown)


src/state/sagas/scenario/DeleteScenario/DeleteScenario.js
```diff
@@ -4,15 +4,27 @@
...
   try {
     const organizationId = action.organizationId;
     const workspaceId = action.workspaceId;
-    yield call(Api.Scenarios.deleteScenario, organizationId, workspaceId, action.scenarioId);
+    const scenarioId = action.scenarioId;
+
+    const response = yield call(Api.Scenarios.findScenarioById, organizationId, workspaceId, scenarioId);
+    action.scenarioRunId = response.data?.lastRun?.scenarioRunId;
+
+    if (response.data.state === SCENARIO_RUN_STATE.RUNNING) {
```
esasova, line 22: is I'm not mistaken `DATA_INGESTION_IN_PROGRESS `can be a source of trouble too, can we include it in the condition ?

nborde-CSM: I didn't even consider DATA_INGESTION_IN_PROGRESS, I am adding it right now.

cypress/commons/utils/apiUtils.js
```diff
@@ -149,6 +149,28 @@ const interceptLaunchScenario = (stubbingOptions) => {
...
+      const scenarioRunId = scenarioRun[1];
+      const scenarioId = stub.getScenarioRunById(scenarioRunId);
+
+      const scenarioLastRun = {
+        id: scenarioRunId,
+        endTime: null,
+        message: null,
+        estimatedDuration: null,
+      };
+      stub.patchScenario(scenarioId, { state: 'Failed', lastRun: scenarioLastRun });
```
csm-thu, line 166: Something's odd here, because `stub.getScenarioRunById(scenarioRunId);`  will return the object of a scenario run, and not a scenario id. I don't think the line `stub.patchScenario(scenarioId [...]` will work.

cypress/commons/utils/apiUtils.js
```diff
@@ -149,6 +149,28 @@ const interceptLaunchScenario = (stubbingOptions) => {
...
+
+      req.reply(scenarioRun);
```
csm-thu, line 168: `req.reply` should be provided with the object of the stubbed _scenario run_ to return, but you're providing the result of the regex `match` here

cypress/e2e/brewery/StopScenarioRun.cy.js
```diff
@@ -34,4 +46,18 @@ describe('can cancel simulation run', () => {
...
+    Scenarios.createScenario(secondScenarioName, true, DATASET.BREWERY_ADT, RUN_TEMPLATE.BREWERY_PARAMETERS).then(
+      (value) => {
+        const scenarioId = value.scenarioCreatedId;
+        ScenarioParameters.launch({ runOptions, saveAndLaunch: true });
+        ScenarioManager.switchToScenarioManager();
+        ScenarioManager.getScenarioAccordion(scenarioId).click();
+        ScenarioManager.getScenarioRunStatus(scenarioId, SCENARIO_STATUS.RUNNING);
+      }
+    );
+    ScenarioManager.deleteScenario(secondScenarioName, true);
```
csm-thu, line 61: Is this line outside of the `createScenario` block on purpose?

csm-thu (changes requested):

cypress/commons/utils/apiUtils.js
nborde-CSM: I would tell _scenarioLastRun_ instead of scenarioRun...

cypress/e2e/brewery/StopScenarioRun.cy.js
nborde-CSM: In other sens, there is no interest to let it inside the block, so I let it outside.

cypress/commons/utils/apiUtils.js
nborde-CSM: In fact, this part of code is useless. I will remove it.

csm-thu (approved):
LGTM :+1:

PR #664, opened 2023-09-20
fix: fix possible read of null data

csm-thu:
The webapp could crash when switching back to the workspaces view while the Scenario Parameters state was dirty.


nborde-CSM (approved):
LGTM :+1:

PR #666, opened 2023-09-21
fix: [PROD-11887] clearing errors in table edit no longer causes dirt…

nborde-CSM:
…y state

- [ ] Code is clean


csm-thu (changes requested):
We need to list all cases this bugfix should handle. See details in ticket PROD-11887

PR #673, opened 2023-10-03
Thu/fix error on file download prod 12416

csm-thu:


cypress/commons/utils/fileUtils.js
```diff
@@ -0,0 +1,24 @@
...
+  formParts.pop();
+  formParts.shift();
+  formParts = formParts.map((el) => (el = el.replace(/\r\n$/, '')));
+  formParts = formParts.map((el) => (el = el.replace(/^\r\n/, '')));
+
+  const form = {};
+  formParts.forEach((formPart) => {
+    const regexMatch = formPart.match(/Content-Disposition: form-data; name="(\w*)"(?:[\S\s]*)?\r\n\r\n([\S\s]*)/);
+    const parameterName = regexMatch[1];
+    const parameterValue = regexMatch[2];
```
esasova, line 16: Just a suggestion: maybe we can use `regexMatch[2]` directly, without assigning it to a variable? Also I'd say that `regexMatch` is too generic, probably something like `parameterByRegex` or `parameterMatch` could be more explicit ? :thinking:

cypress/commons/utils/fileUtils.js
```diff
@@ -0,0 +1,24 @@
+// Copyright (c) Cosmo Tech.
+// Licensed under the MIT license.
+
+const parseMultipartFormData = (dataString) => {
+  const boundary = dataString.split('\r\n')[0];
+  let formParts = dataString.split(boundary);
+  formParts.pop();
+  formParts.shift();
+  formParts = formParts.map((el) => (el = el.replace(/\r\n$/, '')));
+  formParts = formParts.map((el) => (el = el.replace(/^\r\n/, '')));
```
esasova, line 10: Not sure about it but could we simplify this map by `((el) => el.replace(/\r\n$/, ''))` ? :slightly_smiling_face:

esasova (approved):
Just some minor remarks, LGTM :+1:

cypress/commons/utils/fileUtils.js
csm-thu: Oops you're right, this syntax was the leftovers of a first test with `forEach`, I'll change it :slightly_smiling_face:

cypress/commons/utils/apiUtils.js
```diff
@@ -294,6 +295,16 @@ const interceptGetDatasets = () => {
   return alias;
 };
 
+const interceptGetDataset = () => {
+  const alias = forgeAlias('reqGetDataset');
+  cy.intercept({ method: 'GET', url: API_REGEX.DATASET, times: 1 }, (req) => {
+    if (!stub.isEnabledFor('GET_DATASETS')) return;
+    const datasetId = req.url.match(API_REGEX.DATASET)[1];
```
nborde-CSM, line 302: As you made below, maybe add optional chaining operator `?.` before `[1] `

cypress/commons/constants/brewery/IdConstants.js
```diff
@@ -12,6 +12,10 @@ export const BREWERY_SELECTORS = {
         waitersDisabledValue: '[data-cy=nb_waiters-value-disabled-input]',
         waitersInput: '[id=number-input-nb_waiters]',
       },
+      fileUpload: {
+        tabName: '[data-cy=file_upload_tab]',
+        initialStockDataset: '[data-cy=file-upload-initial_stock_dataset]',
```
nborde-CSM, line 17: I dont't realy understand the typo convention. Why don't we write data-cy only with dash `-` ?

cypress/e2e/brewery/FileParameters.cy.js
```diff
@@ -102,6 +108,26 @@ describe('Simple operations on a file parameter', () => {
...
+    Scenarios.createScenario(breweryScenarioName, true, SCENARIO_DATASET, BREWERY_RUN_TEMPLATE).then((value) => {
+      breweryScenarioId = value.scenarioCreatedId;
+      BreweryParameters.switchToFileUploadTab();
+      BreweryParameters.uploadInitialStock(FILE_PATH_1);
+      ScenarioParameters.save();
+      BreweryParameters.getInitialStockFileName().should('have.text', 'CSV file');
+      BreweryParameters.downloadInitialStock();
+      Downloads.checkByContent('initial_stock_dataset.csv', EXPECTED_DATA_AFTER_DUMMY_DATASET_1_UPLOAD);
+
+      route.browse({ url: `${BREWERY_WORKSPACE_ID}/scenario/${breweryScenarioId}` });
```
nborde-CSM, line 125: Is the Is the goal of this action to refresh the page ? In this case, you can use `cy.reload()`

nborde-CSM (approved):
Some questions/remarks but LGMT :+1:

cypress/commons/utils/fileUtils.js
esasova: I'm rather for keeping `parameterName` because of better readability but it's your call :grin:

cypress/commons/constants/brewery/IdConstants.js
esasova: I think it's because of the difference between our data-cy convention (dashes) and id naming convention (underscores), concatenating `data-cy` + `id` gives this thing :slightly_smiling_face:

cypress/e2e/brewery/FileParameters.cy.js
csm-thu: Using `cy.reload()` would not use custom interceptions we have set in the `route.browse` function (e.g. interception of PowerBI token query)

cypress/commons/utils/apiUtils.js
csm-thu: I'm not sure it is really useful though, because it won't help much if the pattern of the dataset id is not detected in the interception url :slightly_smiling_face:

PR #702, opened 2023-11-07
feat: [PROD-12039] implement delete dataset feature

esasova:
- [ ] Code is clean


src/state/sagas/datasets/DeleteDataset/DeleteDataset.js
```diff
@@ -0,0 +1,37 @@
...
+import { DATASET_ACTIONS_KEY } from '../../../commons/DatasetConstants';
+import { STATUSES } from '../../../commons/Constants';
+
+export function* deleteDataset(action) {
+  try {
+    const organizationId = action.organizationId;
+    const datasetId = action.datasetId;
+
+    yield call(Api.Datasets.deleteDataset, organizationId, datasetId);
+    yield call(fetchAllDatasetsData, organizationId);
```
csm-thu, line 18: I'm not a huge fan of fetching all datasets again after deleting one of them, it will have a noticeable impact on perfs :sweat_smile: 
If we change the `DELETE_DATASET` action to remove items from the datasets list, we no longer need to fetch the possible thousands of datasets after each dataset removal.

Note: I think we're doing the same when deleting scenarios, but we probably don't have to ("c'est historique" :grin: )

src/views/DatasetManager/components/DatasetList/DatasetList.js
```diff
@@ -32,8 +31,20 @@ export const DatasetList = () => {
   const { t } = useTranslation();
   const classes = useStyles();
   const [isRefreshConfirmationDialogOpen, setIsRefreshConfirmationDialogOpen] = useState(false);
-  const { sortedDatasetList, currentDataset, selectDataset } = useDatasetList();
+  const { sortedDatasetList, currentDataset, selectDataset, deleteDataset } = useDatasetList();
   const [displayedDatasetList, setDisplayedDatasetList] = useState(sortedDatasetList);
+  const selectedDatasetBeforeDelete = useRef(null);
+
+  useEffect(() => {
+    setDisplayedDatasetList(sortedDatasetList);
```
csm-thu, line 39: Do we really need to update the `displayedDatasetList` state when the "currently selected dataset" changes?

src/views/DatasetManager/components/DatasetList/DatasetList.js
```diff
@@ -149,10 +140,10 @@ export const DatasetList = () => {
       <Card variant="outlined" square={true} sx={{ backgroundColor: 'transparent', mt: 1 }}>
         <List subheader={datasetListHeader}>
           <Divider />
-          {displayedDatasetList.map((dataset) => (
+          {displayedDatasetList.map((dataset, index) => (
             <ListItemButton
               key={dataset.id}
-              selected={dataset.id === currentDataset?.id}
+              selected={currentDataset?.id ? dataset.id === currentDataset.id : index === 0}
```
csm-thu, line 146: This change could lead to the dataset list showing that a dataset is selected although no scenario is selected in the "current dataset" redux reducer (meaning that the dataset overview and dataset metadata panels will be empty).

Can we keep the previous condition `dataset.id === currentDataset?.id` here or does it create problems?

The previous version of the code was selecting the default dataset in the DatasetList hooks (`if (currentDataset == null) selectDataset(sortedDatasetList?.[0] ?? null);`), we could restore this line, or move it to another module if you think it is in the wrong place (for instance, we could move this logic in the saga that retrieves all datasets)

src/views/DatasetManager/components/DatasetList/DatasetList.js
```diff
@@ -108,10 +98,11 @@ export const DatasetList = () => {
       };
       const result = await TwoActionsDialogService.openDialog(dialogProps);
       if (result === 2) {
-        // TODO deleteDataset(datasetId);
+        if (currentDataset?.id !== dataset?.id) selectedDatasetBeforeDelete.current = currentDataset;
```
csm-thu, line 101: I'm not sure why we'd need to store the id of the selected dataset, does the "delete" button change the selected dataset?

csm-thu (changes requested):

src/state/sagas/datasets/DeleteDataset/DeleteDataset.js
```diff
@@ -0,0 +1,51 @@
...
+import { DATASET_ACTIONS_KEY } from '../../../commons/DatasetConstants';
+import { STATUSES } from '../../../commons/Constants';
+
+const getDatasets = (state) => state.dataset.list.data;
+const getSelectedDatasetIndex = (state) => state.dataset.selectedDatasetIndex?.data;
+
+export function* deleteDataset(action) {
+  const datasets = yield select(getDatasets);
+  const selectedDataset = yield select(getSelectedDatasetIndex);
+  const selectedDatasetId = datasets[selectedDataset]?.id;
```
csm-thu, line 17: Maybe we should move this line inside the `try` block below, to prevent the app from crashing if the index of the selected dataset is invalid somehow?

src/state/hooks/DatasetHooks.js
```diff
@@ -29,7 +31,10 @@ export const useAddDatasetToStore = () => {
 export const useCurrentDataset = () => {
   const datasets = useDatasetListData();
   const selectedDatasetIndex = useSelectedDatasetIndex();
-  return datasets?.[selectedDatasetIndex] ?? null;
+  return (
+    datasets?.[selectedDatasetIndex] ??
+    ResourceUtils.getResourceTree(datasets?.filter((dataset) => dataset.main === true))[0]
```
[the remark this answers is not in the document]
esasova: Yes, we can, it's never too safe :laughing: but I don't think it will raise an error because the components won't render if main list is empty (the placeholder is rendered instead), in my tests of deleting the last dataset, the placeholder was shown as expected. However, it can be useful to have a fallback :+1:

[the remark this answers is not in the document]
csm-thu: Yeah, I fear that the extra safeguard may still be useful, because we don't know where this "generic" hook will be used in the future :sweat_smile:

csm-thu (approved):
LGTM :+1:

PR #721, opened 2023-11-23
docs: improve instance view documentation

csm-thu:


esasova (approved):
LGTM :+1:

doc/instanceVisualization.md
```diff
@@ -439,6 +440,19 @@ App:
...
+
+If you encounter the following error:
+
+> IDX10214: Audience validation failed.
+>
+> Audiences: [...]
+>
+> Did not match: validationParameters.ValidAudience: [...]
+
+then you may need to change the list of allowed audiences in the Authentication panel of your Function App (refer to section 2.2 “Configuration“ of the current page).
```
nborde-CSM, line 453: Upper case on `then` ?

csm-thu: I don't think this is necessary, because `then` is the start of the second half of the sentence that starts with `If you encounter the following error:`

PR #726, opened 2023-11-28
test: add selectors & test for datasets list

csm-thu:


cypress/commons/actions/generic/DatasetManager.js
```diff
@@ -23,6 +23,32 @@ function getCreateDatasetButton() {
   return cy.get(GENERIC_SELECTORS.datasetmanager.createDatasetButton);
 }
 
+function getDatasetSearchBar() {
+  return cy.get(GENERIC_SELECTORS.datasetmanager.list.searchBar);
+}
+function getDatasetsList() {
+  return cy.get(GENERIC_SELECTORS.datasetmanager.list.container);
+}
+function findDatasetsListItemButtons(tableParameterElement) {
```
esasova, line 32: could we rename this function to `getDatasetsListItemButtons` ? Actual name sticks better to used cypress function but I found it confusing because it does the same thing that all `get` functions and we don't need to really emphasize the `find` (also, in other generic functions, we don't make difference between `get` and `find`)

esasova (approved):
LGTM :+1:

cypress/commons/actions/generic/DatasetManager.js
csm-thu: Good idea, `get` and `find` are actually more similar than I thought, we probably don't need to make the distinction in our actions names :slightly_smiling_face:

nborde-CSM (approved):
LGTM :+1:

PR #739, opened 2023-12-07
feat: upload twingraph with fetch API

esasova:
- [ ] Code is clean

### DIFF
--- a/src/state/sagas/datasets/CreateDataset/CreateDataset.js
+++ b/src/state/sagas/datasets/CreateDataset/CreateDataset.js
@@ -1,11 +1,34 @@
 // Copyright (c) Cosmo Tech.
 // Licensed under the MIT license.
 
-import { put, takeEvery, call } from 'redux-saga/effects';
+import { call, put, takeEvery } from 'redux-saga/effects';
 import { t } from 'i18next';
+import { Auth } from '@cosmotech/core';
 import { Api } from '../../../../services/config/Api';
 import { DATASET_ACTIONS_KEY } from '../../../commons/DatasetConstants';
 import { dispatchSetApplicationErrorMessage } from '../../../dispatchers/app/ApplicationDispatcher';
+import { TWINGRAPH_STATUS } from '../../../../services/config/ApiConstants';
+
+const uploadZipWithFetchApi = async (organizationId, datasetId, file) => {
+  try {
+    const tokens = await Auth.acquireTokens();
+    const headers = {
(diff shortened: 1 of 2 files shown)


src/state/sagas/datasets/CreateDataset/CreateDataset.js
```diff
@@ -23,13 +46,23 @@ export function* createDataset(action) {
...
 
-    if (dataset.sourceType !== 'None') {
+    if (!['None', 'File'].includes(dataset.sourceType)) {
       yield put({
         type: DATASET_ACTIONS_KEY.TRIGGER_SAGA_REFRESH_DATASET,
         organizationId,
         datasetId: data.id,
       });
     }
+    if (dataset.sourceType === 'File') {
```
csm-thu, line 56: Really small improvement, but this could be an `else if` to prevent checking the second condition (line 56) when the first one (line 49) is true :slightly_smiling_face: 

Another way to organize the conditions would be:
```js
if (dataset.sourceType === 'File') {
  // handle File case
} else if (dataset.sourceType !== 'None') {
  // handle ADT and Storage cases
} 
```
but this second version would probably break if we add new source types in the future :sweat_smile:

src/views/DatasetManager/components/CreateDatasetButton/components/CreateDatasetWizard/CreateDatasetWizard.js
```diff
@@ -73,8 +73,9 @@ export const CreateDatasetWizard = ({ open, closeDialog }) => {
 
   const createDatasetAndCloseDialog = useCallback(() => {
     const values = methods.getValues();
-    if (datasetLocation === 'fromScratch') values.sourceType = 'None';
     DatasetsUtils.removeUndefinedValuesBeforeCreatingDataset(values);
+    if (datasetLocation === 'fromScratch') values.sourceType = 'None';
+    if (values.sourceType === DATASET_SOURCE_TYPE.LOCAL_FILE) values.source = null;
```
csm-thu, line 78: Same minor suggestion, the second condition can never be true when the first one is true, so we can replace the second `if` by an `else if` to make it explicit

nborde-CSM (approved):
LGTM :+1:

src/state/sagas/datasets/CreateDataset/CreateDataset.js
esasova: I'll take the first option, just in case :laughing:

PR #742, opened 2023-12-11
test: add stubbed cypress tests for create ADT, delete, refresh and f…

esasova:
…ilter datasets actions

- [ ] Code is clean

### DIFF
--- a/cypress/commons/services/stubbing.js
+++ b/cypress/commons/services/stubbing.js
@@ -245,6 +245,7 @@ class Stubbing {
   addDataset = (newDataset) => this._addResource('datasets', newDataset);
   getDatasetById = (datasetId) => this._getResourceById('datasets', datasetId);
   patchDataset = (datasetId, datasetPatch) => this._patchResourceById('datasets', datasetId, datasetPatch);
+  deleteDatasetByName = (datasetName) => this._deleteResourceByName('datasets', datasetName);
 
   getSolutions = () => this._getResources('solutions');
   setSolutions = (newSolutions) => this._setResources('solutions', newSolutions);
--- a/cypress/commons/utils/apiUtils.js
+++ b/cypress/commons/utils/apiUtils.js
@@ -369,6 +369,25 @@ const interceptRefreshDataset = () => {
   return alias;
 };
 
+const interceptRefreshDatasetAndPollStatus = (datasetId, options) => {
+  const alias = forgeAlias('reqRefreshDatasetAndPollStatus');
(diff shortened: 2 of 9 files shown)


cypress/commons/services/stubbing.js
```diff
@@ -26,6 +26,8 @@ const STUB_TYPES = [
   'PERMISSIONS_MAPPING',
   'UPDATE_DATASET',
   'UPDATE_SCENARIO',
+  'DELETE_DATASET',
+  'REFRESH_DATASET',
```
csm-thu, line 30: We probably don't need to create custom stubbing types for _delete_ & _refresh_ events, we can enable the stubbing for these events when the stubbing is enabled for `CREATE_DATASET` :slightly_smiling_face: (it could help reduce the number of stubbing options to set when writing new tests)

cypress/commons/utils/apiUtils.js
```diff
@@ -390,6 +409,15 @@ const interceptUpdateDataset = (options) => {
   return alias;
 };
 
+const interceptDeleteDataset = (datasetName) => {
+  const alias = forgeAlias('reqDeleteDataset');
+  cy.intercept({ method: 'DELETE', url: API_REGEX.DATASET, times: 1 }, (req) => {
+    if (stub.isEnabledFor('GET_DATASETS')) stub.deleteScenarioByName(datasetName);
```
csm-thu, line 415: `delete*Scenario*ByName` : this is (probably) not the method you're looking for :sweat_smile: 

Also, the condition should also include ` || stub.isEnabledFor('DELETE_DATASET')` to remove the fake dataset from the data in the `stubbing` class (as mentioned in a previous comment, we can probably use `CREATE_DATASET` instead of creating a new type `DELETE_DATASET`)

csm-thu (approved):
Some suggestions of changes but LGTM :+1:

cypress/commons/services/stubbing.js
esasova: yes, you are right, I introduced `REFRESH_DATASET` to make a difference between refreshing alone and after creating, but finally, I made a special function for the first one, so no need to add an extra stubbing option :+1:

cypress/commons/utils/apiUtils.js
csm-thu: The test might not try to access the stubbed data again. As long as the stubbed "delete" request returns a positive answer, the front-end believes that the dataset is removed :shrug:

nborde-CSM (approved):
LGTM :+1:

PR #749, opened 2023-12-14
Thu/tests dataset overview prod 12475

csm-thu:


src/utils/WorkspacesUtils.js
```diff
@@ -55,6 +67,69 @@ const patchWorkspaceWithDatasetManagerConfiguration = (workspace) => {
...
+      else if (allQueriesIds.includes(query.id))
+        logWarning(`in queries, item #${index} uses a query id that already exists (id: ${query.id}).`);
+      else allQueriesIds.push(query.id);
+
+      if (query.query == null) logWarning(`in queries, item #${index} (id: ${query?.id}) has no query statement`);
+    });
+  }
+
+  const allKpisIds = [];
+  const isQueriesValid = queries != null && queries instanceof Array;
```
esasova, line 96: I think we can check only `queries instanceof Array` because null can't be one :slightly_smiling_face:

src/utils/WorkspacesUtils.js
```diff
@@ -55,6 +67,69 @@ const patchWorkspaceWithDatasetManagerConfiguration = (workspace) => {
...
+    return;
+  }
+
+  const { graphIndicators, categories, queries } = config;
+
+  if (queries != null && !(queries instanceof Array)) {
+    logWarning('property "queries" must be an Array');
+  } else {
+    const allQueriesIds = [];
+    queries.forEach((query, index) => {
```
esasova, line 85: I'm a bit confused : is `queries == null`, the `else` block will be executed and it could provoke a crash on `forEach` ()the same question about `categories` on lines 123-126). Is it a real hypothesis or `queries` cannot be `null` ? :slightly_smiling_face:

esasova (approved):
just some questions but LGTM :+1:

cypress/e2e/brewery/DatasetManager_e2e.cy.js
```diff
@@ -0,0 +1,94 @@
...
+    DatasetManager.startDatasetCreation();
+    DatasetManager.setNewDatasetName(storageDatasetName);
+    DatasetManager.addNewDatasetTag('storage');
+    DatasetManager.setNewDatasetDescription('Dataset created from Storage, during cypress end-to-end test');
+    DatasetManager.getDatasetCreationNextStep().click();
+
+    DatasetManager.selectNewDatasetFromExistingData();
+    DatasetManager.getDatasetCreationNextStep().click();
+
+    DatasetManager.getNewDatasetSourceTypeSelect().click();
```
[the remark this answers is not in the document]
csm-thu: Yes, the removal of the second step is planned in PROD-12831. I hope we deliver this refactoring soon (before the release), so we probably don't need to create an action to factorize these duplicated lines :slightly_smiling_face:

cypress/e2e/brewery/DatasetManager_e2e.cy.js
```diff
@@ -0,0 +1,94 @@
...
+    DatasetManager.getAllRefreshDatasetSpinners().should('be.visible');
+    DatasetManager.getAllRefreshDatasetSpinners(90).should('not.exist');
+    DatasetManager.getIndicatorCard('bars_count').should('be.visible');
+    DatasetManager.getKpiValue(DatasetManager.getIndicatorCard('bars_count')).should('have.text', 1);
+
+    // PROD-12836: previous search filter gets ignored after a change of dataset status, we need to type the same filter
+    // again
+    DatasetManager.getDatasetSearchBarInput().click().clear();
+    DatasetManager.getDatasetSearchBarInput().click().type(storageDatasetName);
+    DatasetManager.getDatasetsListItemButtons().should('have.length', 1);
```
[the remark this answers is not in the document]
csm-thu: Yes, I agree we shouldn't need to check the length of the list, it's not the responsibility of this test.
Yet, I added this extra check for safety reasons:
- this test does not use stubbed resources
- the service account has permissions to remove datasets
- **the dataset manager shows all the datasets of the organization**
When this test runs in our dev environment, there is a risk to delete some actual datasets that we want to keep, so I prefer to keep an extra check :sweat_smile: 

That being said, we have several fail-safe mechanisms:
- `DatasetManager.getAllDeleteDatasetButtons().click();` will raise an error if several buttons are returned
- we check that the dialog obody contains the dataset name
The first one is redundant with the length-check of the list, so I guess it's a good idea to remove it :smile:

src/utils/WorkspacesUtils.js
csm-thu: You're right, thanks :slightly_smiling_face:

src/utils/WorkspacesUtils.js
csm-thu: That's correct, it shows that I should probably write a unit test for this function too :smile:

src/utils/WorkspacesUtils.js
```diff
@@ -71,62 +71,65 @@ const checkDatasetManagerConfiguration = (workspace) => {
...
-    queries.forEach((query, index) => {
-      if (query.id == null) logWarning(`in queries, item #${index} has no id`);
-      else if (allQueriesIds.includes(query.id))
-        logWarning(`in queries, item #${index} uses a query id that already exists (id: ${query.id}).`);
-      else allQueriesIds.push(query.id);
-
-      if (query.query == null) logWarning(`in queries, item #${index} (id: ${query?.id}) has no query statement`);
-    });
+  const isQueriesValid = queries instanceof Array;
+  if (queries != null) {
```
nborde-CSM, line 82: In the case queries is _null_ maybe we should put an additional logWarning in a _else_ block.

nborde-CSM (approved):
One suggestion but LGTM :+1:

PR #766, opened 2023-12-26
refactor: [PROD-12821] replace spinners by skeletons in dataset overview

esasova:
- [ ] Code is clean

### DIFF
--- a/src/views/DatasetManager/components/DatasetOverview/components/CategoryAccordion/components/KPI/KPI.js
+++ b/src/views/DatasetManager/components/DatasetOverview/components/CategoryAccordion/components/KPI/KPI.js
@@ -4,9 +4,10 @@
 import React, { useMemo } from 'react';
 import { useTranslation } from 'react-i18next';
 import PropTypes from 'prop-types';
-import { Grid, Typography } from '@mui/material';
+import { Grid, Skeleton, Typography } from '@mui/material';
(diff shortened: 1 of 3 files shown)


src/views/DatasetManager/components/DatasetOverview/components/CategoryAccordion/components/KPI/KPI.js
```diff
@@ -34,7 +35,11 @@ const KPI = (props) => {
       <Grid item>{label}</Grid>
 
       <Grid item>
-        <KPIValue kpi={kpi} valueTypographyProps={{ ...valueProps }} />
+        {kpi.state === KPI_STATE.IDLE || kpi.state === KPI_STATE.LOADING ? (
+          <Skeleton width="24px" />
+        ) : (
+          <KPIValue kpi={kpi} valueTypographyProps={{ ...valueProps }} />
+        )}
```
csm-thu, line 42: Minor change suggestion: instead of implementing this condition in the `KPI` component, we can implement this logic in the `KPIValue` component, and revert the changes in this file. The skeleton would then be returned by `KPIValue` when the state is `IDLE` or `LOADING` (otherwise the current behavior returns the same result as the "UNKNOWN" state).

I'll add a separated _refactor_ commit for this change and merge the branch, so we can still discuss and revert it later if we prefer to keep the current version :slightly_smiling_face:

PR #768, opened 2023-12-27
fix: [PROD-12832] fix next button disabled in creating wizard if thir…

esasova:
…d step is reached

- [ ] Code is clean


csm-thu:
Just added a commit to fix a minor bug that was hidden by the disabled "Next" button.
The default value of the source type selector in third step was always resetting it to AzureStorage when going from step 2 to step 3, but the content of the third step was not updated, leading to inconsistent states when selecting a different source type in step 3 and then going back to step 2 and forward to step 3 again.

csm-thu (approved):
LGTM :+1:

PR #785, opened 2024-01-11
Thu/fix status and indicators bugs prod 12467

csm-thu:


public/locales/en/translation.json
```diff
@@ -482,6 +482,7 @@
       "datasetNotCreated": "Dataset hasn't been created",
       "datasetNotUpdated": "Dataset hasn't been updated",
       "twingraphNotCreated": "A problem occurred during twingraph creation",
+      "twingraphStatusPollingFailed": "A problem occurred when querying the status of dataset {{datasetDetails}}",
```
nborde-CSM, line 485: English and french translations seem to not be exactly the same. Maybe use "when updating the status" instead of "querying" ?

nborde-CSM (approved):
One question but LGTM :+1:

public/locales/en/translation.json
nborde-CSM: I have no suggestion en français. :relieved:

PR #790, opened 2024-01-16
fix: [PROD-12924] redirect on workspaces selector when user has no ac…

nborde-CSM:
…cess right to the solution

- [ ] Code is clean

### DIFF
--- a/cypress/e2e/brewery/SolutionUnauthorized.cy.js
+++ b/cypress/e2e/brewery/SolutionUnauthorized.cy.js
@@ -1,0 +1,57 @@
+// Copyright (c) Cosmo Tech.
+// Licensed under the MIT license.
+
+import { Login, Workspaces, ErrorBanner } from '../../commons/actions';
+import { stub } from '../../commons/services/stubbing';
+import { GENERIC_SELECTORS } from '../../commons/constants/generic/IdConstants';
+import { EXTENDED_WORKSPACES_LIST } from '../../fixtures/stubbing/default';
+import { apiUtils } from '../../commons/utils';
+
+describe('User has no access to the solution', () => {
+  before(() => {
+    stub.start();
+  });
+
+  after(() => {
+    stub.stop();
+  });
+
+  const solutionError = {
+    statusCode: 403,
+    body: {
+      title: 'Forbidden',
+      status: 403,
+      detail: `RBAC SOL-unauthorized - User does not have permission read`,
+    },
+  };
+
+  // The solution interception is replaced here by a custom one because in this specific case,
+  // we need to return an error instead of the base stubbed solution
+  const customSelectWorkspace = (workspaceId) => {
+    const queries = [apiUtils.interceptPowerBIAzureFunction(), apiUtils.interceptGetScenarios()];
+    Workspaces.getWorkspaceCardById(workspaceId)
+      .should('be.visible')
+      .find(GENERIC_SELECTORS.workspace.openButton)
+      .click();
+    apiUtils.waitAliases(queries, { timeout: 60 * 1000 });
+  };
+
+  it('Redirects to workspace selector when there are several workspaces', () => {
+    stub.setWorkspaces(EXTENDED_WORKSPACES_LIST);
+    const interceptionURL = new RegExp('^' + '/.*/solutions/((sol|SOL)-[\\w]+)');
+    cy.intercept({ method: 'GET', url: interceptionURL, times: 1 }, (req) => {
+      req.reply(solutionError);
+    });
+    Login.login();
+    Workspaces.getWorkspacesView().should('exist');
+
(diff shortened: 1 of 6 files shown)


src/components/UserStatusGate/UserStatusGate.js
```diff
@@ -27,6 +32,12 @@ export const UserStatusGate = ({ children }) => {
     ) : (
       <Navigate to="/accessDenied" replace />
     )
+  ) : solutionUnauthorized ? (
+    location.pathname === '/workspaces' ? (
+      <Workspaces />
+    ) : (
+      <Navigate to="/workspaces" replace />
+    )
```
csm-thu, line 40: Nested ternary operators gets really difficult to read, can we try to replace them by standard `if`blocks ? (early `return` statements in these `if` blocks will prevent nested conditions)

src/state/sagas/solution/FindSolutionById/FindSolutionByIdData.js
```diff
@@ -7,24 +7,41 @@ import { STATUSES } from '../../../commons/Constants';
...
+      type: SOLUTION_ACTIONS_KEY.SET_CURRENT_SOLUTION,
+      status: STATUSES.SUCCESS,
+      solution: data,
+    });
+  } catch (error) {
+    console.error(error);
+    yield put(
+      dispatchSetApplicationErrorMessage(
+        error,
+        t('commoncomponents.banner.', "You don't have permission to access this solution.")
```
csm-thu, line 36: Lack of permissions is probably not the only error that can be caught here (for example, if the workspace configuration has a typo in the solution id, the webapp may try to fetch a solution that doesn't exist).
Can we have a simple distinction between non-existent solutions and permissions errors (based on the return code, for instance), with a default error for other cases: `something went wrong when fetching the solution "<solution id>"`

csm-thu, line 36: The translated key is invalid: `commoncomponents.banner.`
Also, we should probably add the solution id in the error message

nborde-CSM: It was a temporary empty key, while fixing the main problem.

nborde-CSM: It was a temporary empty key, while fixing the main problem.

src/components/UserStatusGate/UserStatusGate.js
nborde-CSM: I did a mix of the two forms

cypress/commons/utils/apiUtils.js
```diff
@@ -424,8 +424,12 @@ const interceptGetSolution = (solutionId) => {
   const alias = forgeAlias('reqGetSolution');
   cy.intercept({ method: 'GET', url: interceptionURL, times: 1 }, (req) => {
     if (!stub.isEnabledFor('GET_SOLUTIONS')) return;
+    // if (stub.solutionError != null) {
+    //  req.reply(stub.solutionError);
+    // } else {
     const solutionIdToGet = solutionId ?? req.url.match(interceptionURL)?.[1];
     req.reply(stub.getSolutionById(solutionIdToGet));
+    // }
```
csm-thu, line 432: Remaining commented code to remove

cypress/e2e/brewery/SolutionUnauthorized.cy.js
```diff
@@ -0,0 +1,55 @@
...
+describe('User has no access to the solution', () => {
+  stub.solutionError = {
+    statusCode: 403,
+    body: {
+      title: 'Forbidden',
+      status: 403,
+      detail: `RBAC SOL-unauthorized - User does not have permission read`,
+    },
+  };
+
```
csm-thu, line 19: There's no need to modify the `stub` object anymore, we can create a `const` to store the object

cypress/e2e/brewery/SolutionUnauthorized.cy.js
```diff
@@ -0,0 +1,55 @@
...
+  before(() => {
+    stub.start();
+  });
+
+  after(() => {
+    stub.stop();
+  });
+
+  const customSelectWorkspace = (workspaceId) => {
+    const queries = [apiUtils.interceptPowerBIAzureFunction(), apiUtils.interceptGetScenarios()];
```
csm-thu, line 29: Maybe we should add a comment here to document that we removed the solution interception from the "usual" queries, and it is replaced by a custom solution interception in the test body

cypress/commons/utils/apiUtils.js
nborde-CSM: Outdated

csm-thu (approved):

esasova (approved):
just one little typo, LGTM :+1:

PR #799, opened 2024-01-25
Thu/option to disable out of sync warning prod 13001

csm-thu:


src/views/Scenario/components/ScenarioDashboardCard/ScenarioDashboardCardHook.js
```diff
@@ -26,16 +28,23 @@ export const useScenarioDashboardCard = () => {
   );
 
   const isDashboardSync = useMemo(() => {
-    if (!currentScenarioRunStartTime) return true;
+    const disableOutOfSyncWarningBanner = (workspace?.webApp?.options?.disableOutOfSyncWarningBanner ?? false) === true;
```
esasova, line 31: In which case we could need this nullish coalescing operator ? if `disableOutOfSyncWarningBanner` isn't defined in the workspace, `undefined === true` will return false, no ? :thinking:

esasova (approved):
One question but LGTM :+1:

src/views/Scenario/components/ScenarioDashboardCard/ScenarioDashboardCardHook.js
csm-thu: Oops, you're right, I'll remove it :smile:  Thank you !

nborde-CSM (approved):
LGTM :+1:

PR #805, opened 2024-02-05
fix: fix dataset disappearing on workspace switch

csm-thu:


src/state/reducers/workspace/WorkspaceReducer.js
```diff
@@ -33,10 +33,19 @@ const workspacesListInitialState = {
 };
 
 const workspacesListReducer = createReducer(workspacesListInitialState, (builder) => {
-  builder.addCase(WORKSPACE_ACTIONS_KEY.SET_ALL_WORKSPACES, (state, action) => {
-    state.data = action.list;
-    state.status = action.status;
-  });
+  builder
+    .addCase(WORKSPACE_ACTIONS_KEY.LINK_TO_DATASET, (state, action) => {
+      state.data.forEach((workspace) => {
```
nborde-CSM, line 38: Is it not more simple to use `find` function ?

`workspaceToLink = state.data.find((workspace) => workspace.id === action.workspaceId;`
`const linkedDatasets = workspaceToLink.linkedDatasetIdList;`

nborde-CSM (approved):
One question but LGMT :+1:

src/state/reducers/workspace/WorkspaceReducer.js
csm-thu: Good idea, I'll make the change :+1:

src/state/reducers/workspace/WorkspaceReducer.js
```diff
@@ -33,10 +33,18 @@ const workspacesListInitialState = {
...
-    state.data = action.list;
-    state.status = action.status;
-  });
+  builder
+    .addCase(WORKSPACE_ACTIONS_KEY.LINK_TO_DATASET, (state, action) => {
+      if (action.datasetId == null) return;
+
+      const targetWorkspace = state.data.find((workspace) => workspace.id === action.workspaceId);
+      const linkedDatasets = targetWorkspace?.linkedDatasetIdList;
+      if (linkedDatasets !== undefined) linkedDatasets.push(action.datasetId);
```
esasova, line 42: just in case: `linkedDatasets` cannot be `null` ?

esasova (approved):
LGTM :+1:

src/state/reducers/workspace/WorkspaceReducer.js
csm-thu: You're right, I think it can be `null` for "legacy" workspaces, I'll replace the condition by `linkedDatasets != null` to handle it correctly. Thank you ! :smile:

PR #810, opened 2024-02-15
Thu/guess missing parameter value var types

csm-thu:


cypress/commons/utils/apiUtils.js
```diff
@@ -625,7 +625,7 @@ const interceptWorkspaceSelectorQueries = () => {
 };
 
 const interceptSelectWorkspaceQueries = () => {
-  return [interceptPowerBIAzureFunction(), interceptGetScenarios(), interceptGetSolution()];
+  return [interceptPowerBIAzureFunction(), interceptGetSolution(), interceptGetScenarios()];
```
nborde-CSM, line 628: What is the interest to only invert the two functions ?

nborde-CSM (approved):
One question but LGTM ?

PR #815, opened 2024-03-13
fix: change behavior when no datasets match the workspace filter

csm-thu:
The previous behavior was to show all 'main' datasets in organization when the workspace filter had no matching results. Yet, this case can happen frequently, for instance when no datasets have been shared with the webapp user: the new behavior is to show an empty dataset list in this case.


nborde-CSM (approved):
LGTM :+1:

esasova (approved):
LGTM :+1:

PR #819, opened 2024-03-18
Thu/minor bug fixes

csm-thu:


src/utils/SolutionsUtils.js
```diff
@@ -62,7 +62,7 @@ const patchSolutionIfLocalConfigExists = async (originalSolution) => {
 };
 
 const checkParametersValidationConstraintsInSolution = (data) => {
-  const parametersWithConstraint = data?.parameters?.filter((parameter) => parameter.options?.validation);
+  const parametersWithConstraint = data?.parameters?.filter((parameter) => parameter && parameter.options?.validation);
```
esasova, line 65: just a question: optional chaining operator only (parameter?.options?.validation) won't work in this case ? :slightly_smiling_face:

esasova (approved):
LGTM :+1:

src/utils/SolutionsUtils.js
csm-thu: Nice catch, I changed it to use the optional chaining operator. Thank you for the suggestion :pray:

nborde-CSM (approved):
LGTM :+1:

PR #820, opened 2024-03-19
feat: [PROD-13080] implement runner stop feature

nborde-CSM:
- [ ] Code is clean


src/state/sagas/runner/StopRunner/StopRunner.js
```diff
@@ -0,0 +1,49 @@
...
+const getSelectedDataset = (state) => state.dataset?.list?.data[state.dataset?.selectedDatasetIndex];
+
+export function* stopRunner(action) {
+  const selectedDataset = yield select(getSelectedDataset);
+  const runnerId = selectedDataset?.source?.name;
+  const organizationId = action.organizationId;
+  const workspaceId = yield select(getWorkspaceId);
+  const selectedDatasetIndex = yield select(getSelectedDatasetIndex);
+
+  if (selectedDataset.ingestionStatus === INGESTION_STATUS.PENDING) {
```
csm-thu, line 22: We're trying to read properties of objects (and arrays)  that can be undefined in some (not common) cases, can we move this part of code inside the _try/catch_ block to make sure we don't crash the whole webapp if these uncommon cases happen?

src/views/DatasetManager/components/DatasetOverview/components/DatasetOverviewPlaceholder/DatasetOverviewPlaceholder.js
```diff
@@ -94,12 +95,23 @@ export const DatasetOverviewPlaceholder = () => {
...
       <ReuploadFileDatasetButton datasetId={currentDatasetId} iconButton={false} />
+    ) : currentDatasetIngestionStatus === INGESTION_STATUS.PENDING ? (
+      <Button
+        data-cy="dataset-overview-abort-button"
+        variant="contained"
+        color="error"
+        onClick={() => stopRunner(currentDatasetId)}
+      >
+        {t('commoncomponents.datasetmanager.overview.placeholder.abortButton', 'Abort')}
+      </Button>
```
csm-thu, line 114: IIUC, the button will be visible for all datasets with status "pending", even those that are not of ETL/runners type. What happens if we try to call `stopRunner` for a dataset of type ADT (or Azure Storage)? 
Shouldn't we only show the "stop runner" button for datasets of type "ETL"?

src/state/sagas/runner/StopRunner/StopRunner.js
esasova: I'd say this condition is a bit redundant because the button is already disabled for all other statuses. If it's somehow necessary, I think it's better to check it in the function that calls the saga in the component, not in the saga itself

src/views/DatasetManager/components/DatasetOverview/components/DatasetOverviewPlaceholder/DatasetOverviewPlaceholder.js
```diff
@@ -94,12 +95,23 @@ export const DatasetOverviewPlaceholder = () => {
...
       return null;
 
     return currentDatasetType === DATASET_SOURCE_TYPE.LOCAL_FILE ? (
       <ReuploadFileDatasetButton datasetId={currentDatasetId} iconButton={false} />
+    ) : currentDatasetIngestionStatus === INGESTION_STATUS.PENDING ? (
+      <Button
+        data-cy="dataset-overview-abort-button"
+        variant="contained"
+        color="error"
+        onClick={() => stopRunner(currentDatasetId)}
```
esasova, line 111: just a question : here, the hook uses `currentDatasetId` as argument but it isn't used elsewhere in the dispatcher or in the saga; on the other hand, the hook takes as argument `runnerId` that we never passed and then, in the saga, both values are calculated from the state :thinking: I must miss something, could you explain how is it wired ? :slightly_smiling_face:

src/state/sagas/runner/StopRunner/StopRunner.js
```diff
@@ -0,0 +1,49 @@
...
+// Licensed under the MIT license.
+import { t } from 'i18next';
+import { put, takeEvery, call, select } from 'redux-saga/effects';
+import { Api } from '../../../../services/config/Api';
+import { INGESTION_STATUS } from '../../../../services/config/ApiConstants';
+import { DATASET_ACTIONS_KEY } from '../../../commons/DatasetConstants';
+import { RUNNER_ACTIONS_KEY } from '../../../commons/RunnerConstants';
+import { dispatchSetApplicationErrorMessage } from '../../../dispatchers/app/ApplicationDispatcher';
+
+const getWorkspaceId = (state) => state.workspace.current?.data?.id;
```
esasova, line 11: IIUC, we pass the `workspaceId` via the hook, do we really need to recalculate it in the saga ? :thinking:

nborde-CSM: Right, it is a mess...

src/views/DatasetManager/components/DatasetOverview/components/DatasetOverviewPlaceholder/DatasetOverviewPlaceholder.js
nborde-CSM: It's a mess

src/views/DatasetManager/components/DatasetOverview/components/DatasetOverviewPlaceholder/DatasetOverviewPlaceholder.js
nborde-CSM: Yes, I should have removed this parameter after having made changes in the saga.

src/state/sagas/runner/StopRunner/StopRunner.js
```diff
@@ -0,0 +1,46 @@
...
+      type: DATASET_ACTIONS_KEY.UPDATE_DATASET,
+      datasetId,
+      datasetData: { ingestionStatus: INGESTION_STATUS.ERROR },
+    });
+  } catch (error) {
+    console.error(error);
+    yield put(
+      dispatchSetApplicationErrorMessage(
+        error,
+        t('views.runner.runnerStopError.comment', 'Could not stop runner "{{runnerId}}".', {
```
csm-thu, line 34: Isn't the key `views.runner.runnerStopError.comment` missing from translation files?

csm-thu (approved):
LGTM :+1:

esasova (approved):
LGTM :+1:

PR #827, opened 2024-04-03
thu/subdataset prod 13195

csm-thu:


src/views/DatasetManager/components/CreateDatasetButton/CreateSubDatasetButtonHook.js
```diff
@@ -0,0 +1,42 @@
...
+
+  const dataSourceRunTemplates = useMemo(() => {
+    const parameters = solutionData.parameters;
+    const runTemplatesParameters = solutionData.runTemplatesParametersIdsDict;
+    const dataSourcesWithParameters = customSubDataSourceRunTemplates.map((dataSource) => ({
+      ...dataSource,
+      parameters: parameters.filter((parameter) => runTemplatesParameters[dataSource.id].includes(parameter.id)),
+    }));
+
+    const runTemplates = {};
```
nborde-CSM, line 21: Is it necessary to set `runtemplate` as _object null_ instead of _null_ ?

nborde-CSM (approved):
One question but LGTM :+1:

src/views/DatasetManager/components/CreateDatasetButton/CreateSubDatasetButtonHook.js
csm-thu: Yes, because the next line tries to create keys in this object, and we can't do that if it's `null`:
`   [...]    runTemplates[runTemplate.id] = runTemplate));`

nborde-CSM (approved):
LGTM :+1:

PR #828, opened 2024-04-05
feat: place delete button in overview area also

sellisd:
- [ ] Code is clean

### DIFF
--- a/src/views/DatasetManager/components/DatasetOverview/components/DatasetActions/DatasetActions.js
+++ b/src/views/DatasetManager/components/DatasetOverview/components/DatasetActions/DatasetActions.js
@@ -1,0 +1,15 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import ButtonGroup from '@mui/material/ButtonGroup';
+import { DeleteDatasetButton } from '../../../DeleteDatasetButton/DeleteDatasetButton';
+
+export default function DatasetActions({ dataset }) {
+  return (
+    <ButtonGroup>
+      <DeleteDatasetButton dataset={dataset} location="dataset-actions-" />
+    </ButtonGroup>
+  );
+}
+DatasetActions.propTypes = {
+  dataset: PropTypes.object.isRequired,
+};
--- a/src/views/DatasetManager/components/DeleteDatasetButton/DeleteDatasetButton.js
+++ b/src/views/DatasetManager/components/DeleteDatasetButton/DeleteDatasetButton.js
@@ -10,7 +10,7 @@ import { TwoActionsDialogService } from '../../../../services/twoActionsDialog/t
 import { useDeleteDatasetButton } from './DeleteDatasetButtonHooks';
 
 export const DeleteDatasetButton = ({ dataset, location }) => {
-  const isDisabled = dataset.ingestionStatus === INGESTION_STATUS.PENDING;
+  const isDisabled = !dataset || dataset.ingestionStatus === INGESTION_STATUS.PENDING;
   const { t } = useTranslation();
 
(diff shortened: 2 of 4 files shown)


cypress/e2e/brewery/DatasetManager_e2e.cy.js
```diff
@@ -40,7 +40,7 @@ describe('End-to-end test of the dataset manager view', () => {
     DatasetManager.getIndicatorCard('bars_count').should('be.visible');
     DatasetManager.getKpiValue(DatasetManager.getIndicatorCard('bars_count')).should('have.text', 1);
 
-    DatasetManager.getAllDeleteDatasetButtons().click();
+    DatasetManager.getAllDeleteDatasetButtons().eq(0).click();
```
csm-thu, line 43: I think using `eq(0)` won't guarantee that we click on the `DatasetList` button, so this change might modify the behavior of the test. We should probably use different selectors or "data-cy" tags to make sure we can select the button we want to test. `getAllDeleteDatasetButtons` was initially only retrieving the buttons in `DatasetList`, maybe we should change the data-cy value of the Overview button so it won't be selected ?

nborde-CSM (approved):
Some remarks but LGTM :+1:

src/views/DatasetManager/components/DeleteDatasetButton/DeleteDatasetButton.js
```diff
@@ -59,7 +59,7 @@ export const DeleteDatasetButton = ({ dataset, location }) => {
     <PermissionsGate userPermissions={userPermissionsOnDataset} necessaryPermissions={[ACL_PERMISSIONS.DATASET.DELETE]}>
       <IconButton
         onClick={(event) => askConfirmationToDeleteDialog(event, dataset)}
-        data-cy={`dataset-delete-button-${location}${dataset.id}`}
+        data-cy={`dataset-delete-button-${location}${dataset?.id}`}
```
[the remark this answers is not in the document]
csm-thu: This is a reusable component, we can't be sure that `dataset` won't be `null` at some point (despite `isRequired` option on the prop). We should probably keep the optional chaining operator :slightly_smiling_face:

cypress/e2e/brewery/DatasetManager_e2e.cy.js
sellisd: The two buttons have different data-cy strings. The one in the dataset list is named `dataset-delete-button-<DATASETID>` and the new on in the top right is called `dataset-delete-button-DatasetActions<DATASETID>` The cypress test fails because it tries to click on both buttons at the same time as it is using the `getAllDeleteDatasetButtons` function which searches for attributes starting with `dataset-delete-button-`. To avoid the  `eq(0)` we could instead modify the data-cy of the delete button by prepending the location string.

src/views/DatasetManager/components/DatasetOverview/components/DatasetActions/DatasetActions.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import ButtonGroup from '@mui/material/ButtonGroup';
+import { DeleteDatasetButton } from '../../../DeleteDatasetButton/DeleteDatasetButton';
+
+export default function DatasetActions({ dataset }) {
+  return (
+    <ButtonGroup>
+      <DeleteDatasetButton dataset={dataset} location="DatasetActions" />
```
csm-thu, line 9: Could we replace `location="DatasetActions"` by `location="dataset-actions-"`? This would make the resulting data-cy more consistent with other data-cy values (`dataset-actions-dataset-delete-button-<dataset_id>`)

csm-thu (approved):
One improvement suggestion, but LGTM :+1:

PR #830, opened 2024-04-08
fix: fix runner logs parsing to keep only logs of main container

csm-thu:
This commit fixes an error during runner logs download when the number of streams in logs was lower than 2 (e.g. when stderr was empty), and fixes undesired logs (i.e. not from main container) in the downloaded file when the order of streams in the API response was not the order expected by the webapp.


nborde-CSM (approved):
LGTM :+1:

esasova (approved):
LGTM :+1:

PR #860, opened 2024-07-03
feat: [SDCOSMO-1974] implement scenario selector for scenario parameters

nborde-CSM:
- [ ] Code is clean

### DIFF
--- a/cypress/commons/actions/brewery/BreweryParameters.js
+++ b/cypress/commons/actions/brewery/BreweryParameters.js
@@ -128,6 +128,20 @@ const getCommentInput = () => ScenarioParameters.getParameterInput('text-input-c
 const getAdditionalDate = () => ScenarioParameters.getParameterValue('date-input-additional_date');
 const getAdditionalDateHelperText = () => cy.get('[id=date-text-field-additional_date-helper-text]');
 const getAdditionalDateInput = () => ScenarioParameters.getParameterInput('date-input-additional_date');
+const getScenarioToCompare = () => ScenarioParameters.getParameterValue('single-select-scenario_to_compare');
+
+const getScenarioToCompareSelect = () => {
+  return cy.get(BREWERY_SELECTORS.scenario.parameters.additionalParameters.scenarioToCompare);
+};
+const getScenarioToCompareSelectInput = () => {
+  return ScenarioParameters.getParameterInput('single-select-scenario_to_compare');
+};
+const selectScenarioToCompareOption = (option) => {
+  getScenarioToCompareSelect().click();
+  cy.get(
+    BREWERY_SELECTORS.scenario.parameters.additionalParameters.scenarioToCompareOption.replace('$OPTION', option)
+  ).click();
+};
 
 function getCurrencyTextField() {
   return getCurrencyParameterContainer().find(GENERIC_SELECTORS.genericComponents.basicEnumInput.textField);
@@ -574,6 +588,10 @@ export const BreweryParameters = {
   getAdditionalTablesInput,
   getCommentInput,
   getAdditionalDateInput,
+  getScenarioToCompareSelect,
+  getScenarioToCompareSelectInput,
+  getScenarioToCompare,
+  selectScenarioToCompareOption,
   getCurrencyNameHelperText,
   getEvaluationHelperText,
   getCommentHelperText,
--- a/cypress/e2e/brewery/ScenarioParameters-additional-advanced.cy.js
+++ b/cypress/e2e/brewery/ScenarioParameters-additional-advanced.cy.js
@@ -40,6 +40,8 @@ describe('Additional advanced scenario parameters tests', () => {
       additionalTables: 3,
       comment: 'None',
       additionalDate: '06/22/2022',
+      countries: [],
+      scenarioToCompare: '',
     };
     const VALUES_TO_UPDATE = {
       additionalSeats: 888,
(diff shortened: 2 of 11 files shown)


src/components/ScenarioParameters/components/ScenarioParametersInputs/ScenarioSelect.js
```diff
@@ -0,0 +1,68 @@
...
+import { Select } from '@cosmotech/ui';
+import { useSortedScenarioList } from '../../../../hooks/ScenarioListHooks';
+import { useCurrentScenarioId } from '../../../../state/hooks/ScenarioHooks';
+import { TranslationUtils } from '../../../../utils';
+
+export const ScenarioSelect = ({ parameterData, context, parameterValue, setParameterValue, isDirty }) => {
+  const { t } = useTranslation();
+  const scenarioList = useSortedScenarioList();
+  const currentScenarioId = useCurrentScenarioId();
+  const runTemplateWhiteList = parameterData.options?.runTemplateFilter;
```
csm-thu, line 16: "whitelist" is a single word, I think there shouldn't be a capital "L" in the variable name
Also, maybe we could simply rename the variable to `runTemplateFilter` to stay consistent with the option name?

src/components/ScenarioParameters/components/ScenarioParametersInputs/ScenarioSelect.js
```diff
@@ -0,0 +1,68 @@
...
+    scenarioList.forEach((scenario) => {
+      const simulationRun = scenario?.lastRun?.csmSimulationRun;
+      if (
+        runTemplateWhiteList.includes(scenario.runTemplateId) &&
+        simulationRun != null &&
+        scenario.id !== currentScenarioId
+      )
+        filteredScenarioTempList.push({ key: simulationRun, label: scenario.name });
+    });
+    return filteredScenarioTempList;
```
esasova, line 30: It's a bit confusing : if there is no `runTemplateWhiteList`, we return the whole `scenarioList` but in the second condition `runTemplateWhiteList` is mandatory. Also, in the first case we return a scenarios object with all properties and in the second, a `key/label` pair (the shape the `Select` component waits) I don't see how it works... :thinking: 

Finally, is it possible to refactor the function with `filter()` and `map()` methods to be more consice ? :pray:

cypress/commons/actions/brewery/BreweryParameters.js
```diff
@@ -128,6 +128,16 @@ const getCommentInput = () => ScenarioParameters.getParameterInput('text-input-c
 const getAdditionalDate = () => ScenarioParameters.getParameterValue('date-input-additional_date');
 const getAdditionalDateHelperText = () => cy.get('[id=date-text-field-additional_date-helper-text]');
 const getAdditionalDateInput = () => ScenarioParameters.getParameterInput('date-input-additional_date');
+const getScenarioToCompareInput = () => ScenarioParameters.getParameterInput('single-select-scenario_to_compare');
+const getScenarioToCompare = () => ScenarioParameters.getParameterValue('single-select-scenario_to_compare');
+
+const getScenarioToCompareSelect = () => {
+  return getScenarioToCompareInput();
+};
```
csm-thu, line 136: Do we need a function that only calls another function?
Also, the 3 lines are equivalent to
`const getScenarioToCompareSelect = getScenarioToCompareInput;`

cypress/commons/actions/brewery/BreweryParameters.js
```diff
@@ -128,6 +128,16 @@ const getCommentInput = () => ScenarioParameters.getParameterInput('text-input-c
...
+const selectScenarioToCompareOption = (option) => {
+  getScenarioToCompareSelect().click();
+  cy.contains(option).click();
```
csm-thu, line 139: Calling `cy.contains(option)` this way could lead to unexpected behavior, because it will return the first element in the DOM containing the `option` value.

Can we use the same mechanism with find/replace as the other select components instead ?

Example:
```js
function getCurrencySelectOption(option) {
  getCurrencySelect().click();
  cy.get(BREWERY_SELECTORS.scenario.parameters.basicTypes.currencyOption.replace('$OPTION', option)).click();
}
```

cypress/e2e/brewery/ScenarioParameters-additional-advanced.cy.js
```diff
@@ -88,6 +94,8 @@ describe('Additional advanced scenario parameters tests', () => {
       .type('{moveToStart}' + '29/08/1997')
       .should('value', '29/08/1997');
 
+    BreweryParameters.selectScenarioToCompareOption(VALUES_TO_UPDATE.scenarioToCompare);
```
csm-thu, line 97: We should probably add a check with `should('have.text', ...` before the _discard_ too

doc/scenarioParametersConfiguration.md
```diff
@@ -757,15 +757,16 @@ components, whose value depend on the **input component type** and the **paramet
...
-| `string`      |          | Text field           | `text-input-<parameterId>`   |
-| `%DATASETID%` |          | File upload          | `file-upload-<parameterId>`  |
-| `%DATASETID%` | `TABLE`  | Editable table       | `table-<parameterId>`        |
+| varType       | subType     | input component      | `data-cy` pattern               |
+| ------------- | ----------- | -------------------- | ------------------------------- |
+| `bool`        |             | Switch boolean input | `toggle-input-<parameterId>`    |
+| `date`        |             | Date picker          | `date-input-<parameterId>`      |
+| `enum`        |             | Dropdown list        | `enum-input-<parameterId>`      |
+| `enum`        | `RADIO`     | Radio buttons        | `radio-input-<parameterId>`     |
+| `enum`        | `SCENARIOS` | Scenarios select     | `scenarios-input-<parameterId>` |
```
csm-thu, line 766: Is the `data-cy` pattern up-to-date? In cypress id constants, we're using `single-select-<parameterId>`

cypress/e2e/brewery/ScenarioParameters-additional-advanced.cy.js
```diff
@@ -88,6 +94,11 @@ describe('Additional advanced scenario parameters tests', () => {
       .type('{moveToStart}' + '29/08/1997')
       .should('value', '29/08/1997');
 
+    BreweryParameters.selectScenarioToCompareOption(VALUES_TO_UPDATE.scenarioToCompare).should(
```
esasova, line 97: I'm not sure about this assertion because the used function `selectScenarioToCompareOption` doesn't return an element to assert for :thinking:

cypress/e2e/brewery/ScenarioParameters-additional-advanced.cy.js
```diff
@@ -105,6 +116,7 @@ describe('Additional advanced scenario parameters tests', () => {
     BreweryParameters.switchToAdditionalParametersTab();
     BreweryParameters.getCommentInput().should('value', INIT_VALUES.comment);
     BreweryParameters.getAdditionalDateInput().should('value', INIT_VALUES.additionalDate);
+    BreweryParameters.getScenarioToCompareInput().should('have.text', '');
```
esasova, line 119: I don't see this function in `BreweryParameters` export (the same on line 119) :thinking:

PR #887, opened 2024-08-01
Thu/add api key support

csm-thu:


src/services/powerbi/PowerBIService.js
```diff
@@ -5,13 +5,19 @@ import { GET_EMBED_INFO_URL } from '../../state/commons/PowerBIConstants';
...
  * Get PowerBI accessible data
  * @return Details like Embed URL, Access token and Expiry, errors if any
  */
 const getPowerBIData = (powerBIWorkspaceId, reportsIds, logInWithUserCredentials) => {
+  if (process.env.REACT_APP_API_KEY)
+    return {
+      accesses: null,
+      error: forgePowerBIError('', 'PowerBI disabled', 'API key detected, PowerBI dashboards have been disabled'),
+    };
+
```
esasova, line 20: If we use the API key, is it a real error ? :thinking: I think it can be confusing for end users to see a red error banner while this connection method can't be configured to use PowerBi. Maybe, a `console.error` for integrators is less intrusive? The chances to get this use case are low because an integrator won't configure powerBi with API key so disabled option will come first but it can happen if the new workspace is a copy of an existing one with powerBi :slightly_smiling_face:

csm-thu: End-users will probably never show this error message, because we must only use API keys with local webapps (otherwise, the API key could leak _via_ the headers of the API queries).
Since only devs will see this error message, I thought a banner would be more straightforward to show why the charts are disabled, even if the cause comes from the configuration and it's not really an error :sweat_smile:

esasova: In this case, we can keep an error banner :slightly_smiling_face:

cypress/commons/actions/generic/Login.js
```diff
@@ -37,9 +37,12 @@ function login(options) {
 
       let reqAuthAlias;
       let browseCallback;
-      if (auth.USE_SERVICE_ACCOUNT) {
+      if (auth.USE_SERVICE_ACCOUNT || auth.USE_API_KEY) {
```
nborde-CSM, line 40: Same question as usual, is there any chance to have error with attribute as _false_ instead of _null/undefined_ ?

nborde-CSM (approved):
One little question but LGTM :+1:

cypress/commons/actions/generic/Login.js
csm-thu: Both constants `USE_SERVICE_ACCOUNT` and `USE_API_KEY` are defined in the _authUtils_ module, they are boolean values returned by `Cypress.env('<some env var>') === 1;`. Unless this code changes, they can't be `null` nor `undefined`

PR #894, opened 2024-08-13
Nbo/overview category details prod 13564

nborde-CSM:
- [ ] Code is clean


src/views/DatasetManager/components/DatasetOverview/components/CategoryAccordion/components/CategoryDetailsDialog/CategoryDetailsDialog.js
```diff
@@ -0,0 +1,157 @@
...
+          ),
+          secondButtonText: t('commoncomponents.banner.button.copy.label', 'Copy'),
+          toggledButtonText: t('commoncomponents.banner.button.copy.copied', 'Copied'),
+        }}
+        clearErrors={clearApplicationErrorMessage}
+      />
+    );
+  }, [applicationError, clearApplicationErrorMessage, t]);
+
+  const categories = useMemo(() => {
```
csm-thu, line 49: The const should probably be renamed from `categories` to `categoryKpis` or something similar :slightly_smiling_face:

src/views/DatasetManager/components/DatasetOverview/components/CategoryAccordion/components/CategoryDetailsDialog/CategoryDetailsDialog.js
```diff
@@ -0,0 +1,157 @@
...
+    </>
+  );
+};
+
+CategoryDetailsDialog.propTypes = {
+  category: PropTypes.object.isRequired,
+  kpis: PropTypes.array,
+};
+
+CategoryDetailsDialog.defaultProps = {};
```
csm-thu, line 157: You're calling the `map` and `find` methods on the `kpi` prop without checking if it's set, maybe we should set an empty array as default value to prevent some errors.

src/state/sagas/datasets/QueryDatasetTwingraph/QueryDatasetTwingraph.js
```diff
@@ -44,8 +44,16 @@ function* runDatasetTwingraphQuery(action, query, attemptsNumber = 0) {
...
   const { workspace } = action.payload;
   const queries = workspace?.webApp?.options?.datasetManager?.queries ?? [];
+  const graph = workspace?.webApp?.options?.datasetManager?.graphIndicators;
+  const categories = workspace?.webApp?.options?.datasetManager?.categories;
+
   yield all(
     queries.map((query, index) => {
+      if (
+        !categories.some((cat) => cat?.kpis.some((kpi) => kpi?.queryId === query.id)) &&
+        !graph.some((g) => g?.queryId === query.id)
```
csm-thu, line 54: Could we rename `g` to something like `kpi`, `indicator` or `graphIndicator`?
Also, `graph` is an array, maybe `graphIndicators` would be a better name for it

src/state/sagas/datasets/QueryDatasetTwingraph/QueryDatasetTwingraph.js
```diff
@@ -44,8 +44,16 @@ function* runDatasetTwingraphQuery(action, query, attemptsNumber = 0) {
...
+      )
+        return null;
```
csm-thu, line 56: Is it okay to have `null` items inside the array of effects provided to `yield all`?
Not sure how it will behave, but the doc mentions:
> When we yield an array of effects, the generator is blocked until all the effects are resolved **or as soon as one is rejected**

We should probably double check that a `null` item won't cancel the other queries :slightly_smiling_face:

src/config/overrides/Workspaces.js
```diff
@@ -228,7 +248,32 @@ export const WORKSPACES = [
                   queryId: 'customers',
                 },
               ],
-              attributes: ['Satisfaction', 'SurroundingSatisfaction', 'Thirsty'],
+              attributes: ['Name', 'Satisfaction', 'Surrounding', 'Thirsty'],
```
csm-thu, line 251: The attribute `SurroundingSatisfaction` was replaced by `Surrounding`: was it done on purpose?

src/config/overrides/Workspaces.js
```diff
@@ -228,7 +248,32 @@ export const WORKSPACES = [
...
+              previewTable: {
+                columns: [
+                  {
+                    field: 'name',
```
csm-thu, line 255: The case of field names is not consistent between bar and customer attributes (`NbWaiters`, `RestockQty`, `Stock` vs. `name`, `satisfaction`, ...). Could we change it to make it more consistent?

src/components/ScenarioParameters/components/ScenarioParametersInputs/GenericTable.js
```diff
@@ -68,7 +66,7 @@ export const GenericTable = ({
   const selectedRowsRef = useRef();
 
   const tableLabels = {
-    label: t(TranslationUtils.getParameterTranslationKey(parameterId), parameterId),
+    label: t(TranslationUtils.getParameterTranslationKey(parameterId), ''),
```
csm-thu, line 69: I think this change would break the default behavior that displays the parameter id as fallback when no "translated labels" are provided :thinking:

src/config/overrides/Workspaces.js
nborde-CSM: It is a mess caused by the multiple refactor of _attributes_

src/state/sagas/datasets/QueryDatasetTwingraph/QueryDatasetTwingraph.js
nborde-CSM: I have tested queries in different orders in the workspace config so as if a null item would cancel the other queries, errors would instantly be visible on the overview kpis.
It's not the case.

src/views/DatasetManager/components/DatasetOverview/components/CategoryAccordion/components/CategoryDetailsDialog/CategoryDetailsDialog.js
```diff
@@ -0,0 +1,162 @@
...
+  );
+};
+
+CategoryDetailsDialog.propTypes = {
+  category: PropTypes.object.isRequired,
+  kpis: PropTypes.array,
+};
+
+CategoryDetailsDialog.defaultProps = {
+  KPIValue: [],
```
csm-thu, line 161: The prop name is probably wrong here, shouldn't it be `kpis` instead of `KPIValue`?

src/components/ScenarioParameters/components/ScenarioParametersInputs/GenericTable.js
```diff
@@ -210,14 +207,14 @@ export const GenericTable = ({
     }
     GenericTable.downloadLocked[lockId] = true;
     try {
-      if (!currentScenarioDatasetList || currentScenarioDatasetList?.length === 0)
+      const sourceDatasetId = context.targetDatasetId;
+      if (!sourceDatasetId)
```
esasova, line 211: If we delete `currentScenarioDatasetList` from the component, _GenericTable_ as scenario parameter won't be able to get the id of its dataset because it doesn't have it in `context` :thinking: we need to keep `currentScenarioDatasetList` as fallback value, otherwise to pass it through `context` as well

src/views/DatasetManager/components/DatasetOverview/components/CategoryAccordion/components/CategoryDetailsDialog/CategoryDetailsDialog.js
```diff
@@ -0,0 +1,162 @@
...
+        parameterData={parameterData}
+        context={context}
+        parameterValue={parameterValue}
+        setParameterValue={() => {}}
+        resetParameterValue={() => {}}
+      />
+    );
+  }, [query, open, isDarkTheme, datasetId, category?.id, category?.previewTable?.columns]);
+
+  return query == null ? null : (
```
esasova, line 115: Maybe React syntax `query && ...` is slightly more readable ?

src/views/DatasetManager/components/DatasetOverview/components/CategoryAccordion/components/CategoryDetailsDialog/CategoryDetailsDialog.js
```diff
@@ -0,0 +1,162 @@
...
+          toggledButtonText: t('commoncomponents.banner.button.copy.copied', 'Copied'),
+        }}
+        clearErrors={clearApplicationErrorMessage}
+      />
+    );
+  }, [applicationError, clearApplicationErrorMessage, t]);
+
+  const categoryKpis = useMemo(() => {
+    return (
+      <Grid container sx={{ p: 5 }} spacing={'64px'} justifyContent="center">
```
esasova, line 54: I think, we can replace _64px_ by `spacing={8}` to avoid to hard-code values

src/services/config/WorkspaceSchema.js
```diff
@@ -142,6 +142,23 @@ const basicWebAppOptions = z.object({
...
+                    .array(
+                      z
+                        .object({ field: z.string(), headerName: z.string(), type: z.array(z.string()) })
+                        .strict()
+                        .optional()
+                        .nullable()
+                    )
+                    .optional()
+                    .nullable(),
+                  queryId: z.string(),
```
esasova, line 157: It's just a question, but why only `queryId` isn't `optional` neither `nullable` ? The component cannot be displayed without columns, so I'd say either both have to be `optional` and `nullable` or none of them :slightly_smiling_face:

src/components/ScenarioParameters/components/ScenarioParametersInputs/GenericTable.js
csm-thu: When used in the _scenario parameters_ panel, I think the context [already has](https://github.com/Cosmo-Tech/azure-sample-webapp/blob/main/src/components/ScenarioParameters/ScenarioParameters.js#L150) the scenario dataset as `targetDatasetId` (IIRC, it's been added in the context for the dynamic values of _enum_ and _multi-select_ components)

src/views/DatasetManager/components/DatasetOverview/components/CategoryAccordion/components/CategoryDetailsDialog/CategoryDetailsDialog.js
nborde-CSM: In this case I need a ternary operator

public/locales/en/translation.json
```diff
@@ -747,24 +747,29 @@
...
-        "notTwingraphDatasetError": "Only twingraph datasets can be used to fetch table data dynamically",
-        "noQueryError": "Impossible to fetch data from dataset because there is no twingraph query defined for this parameter. You can load data manually using Import button",
+        "noQueryError": "Impossible to fetch data from dataset because there is no twingraph query defined for this parameter. You can load data manually using the Import button",
+        "NoQueryErrorImportFalse": "Impossible to fetch data from dataset because there is no twingraph query defined for this parameter.",
         "wrongResultKeyError": "Returned result doesn't have {{resultKey}} property. Probably there is an error in resultKey configuration, please, check your solution",
-        "wrongQueryError": "Returned result is empty, there is probably an error in your query configuration. Please, check your solution"
+        "wrongQueryError": "Returned result is empty, there is probably an error in your query configuration. Please, check your solution",
+        "unknownQueryError": "Something went wrong when querying the table data. Please contact your administrator if the problem persits."
```
esasova, line 772: typo: persi**S**ts :slightly_smiling_face:

public/locales/fr/translation.json
```diff
@@ -747,24 +747,29 @@
...
         "noQueryError": "Il est impossible de charger les données depuis le dataset parce qu'aucune requête du twingraph n'est définie pour ce paramètre. Vous pouvez charger les données manuellement avec le bouton Importer",
+        "NoQueryErrorImportFalse": "Il est impossible de charger les données depuis le dataset parce qu'aucune requête du twingraph n'est définie pour ce paramètre.",
         "wrongResultKeyError": "Le résultat de la requête ne contient pas de propriété {{resultKey}}. Cela provient probablement d'une erreur dans la configuration de resultKey, veuillez vérifier la solution.",
-        "wrongQueryError": "Le résultat retourné est vide. Cela provient probablement d'une erreur dans la configuration de la requête. Veuillez vérifier la solution."
+        "wrongQueryError": "Le résultat retourné est vide. Cela provient probablement d'une erreur dans la configuration de la requête. Veuillez vérifier la solution.",
+        "unknownQueryError": "Something went wrong when querying the table data. Please contact your administrator if the problem persits."
```
esasova, line 772: Missing french translation :slightly_smiling_face:

src/components/ScenarioParameters/components/ScenarioParametersInputs/GenericTable.js
```diff
@@ -67,24 +65,41 @@ export const GenericTable = ({
...
-    loading: t('genericcomponent.table.labels.loading', 'Loading...'),
-    clearErrors: t('genericcomponent.table.button.clearErrors', 'Clear'),
-    errorsPanelMainError: t('genericcomponent.table.labels.fileImportError', 'File load failed.'),
-    placeholderTitle: canChangeRowsNumber
-      ? t('genericcomponent.table.labels.addRowPlaceholderTitle', 'Import or create your first data')
-      : t('genericcomponent.table.labels.placeholderTitle', 'Import your first data'),
-    placeholderBody: canChangeRowsNumber
-      ? t(
+  const [placeholderTitle, setPlaceholderTitle] = useState();
+  const [placeholderBody, setPlaceholderBody] = useState();
```
esasova, line 69: As far as I can see, we never set `placeholderTitle` and `placeholderBody` separately, maybe it's better to use one `useState` for both, with an object with keys `title` and `body` ?

src/components/ScenarioParameters/components/ScenarioParametersInputs/GenericTable.js
```diff
@@ -197,6 +211,16 @@ export const GenericTable = ({
...
   const _getDataFromTwingraphDataset = async (setClientFileDescriptor) => {
+    const _setClientFileDescriptorToNull = () => {
+      setClientFileDescriptor({
+        file: null,
+        content: null,
+        agGridRows: null,
+        errors: null,
+        tableDataStatus: TABLE_DATA_STATUS.ERROR,
+      });
+    };
```
esasova, line 222: this function is used only once, not sure there is a real benefit to create it :thinking:

src/components/ScenarioParameters/components/ScenarioParametersInputs/GenericTable.js
```diff
@@ -209,76 +233,107 @@ export const GenericTable = ({
...
-        setClientFileDescriptor({
-          name: fileName,
-          file: null,
-          agGridRows: agGridData.rows,
-          errors: null,
-          status: UPLOAD_FILE_STATUS_KEY.READY_TO_UPLOAD,
-          tableDataStatus: TABLE_DATA_STATUS.READY,
-          uploadPreprocess: { content: _uploadPreprocess },
-        });
+      ({ data } = await Api.Datasets.twingraphQuery(organizationId, sourceDatasetId, { query }));
```
esasova, line 265: There are parenthesis around the api call, I don't know this syntax, is it some linter games or something new ? :slightly_smiling_face:

src/components/ScenarioParameters/components/ScenarioParametersInputs/GenericTable.js
```diff
@@ -209,76 +233,107 @@ export const GenericTable = ({
...
     } catch (error) {
-      const errorComment = error?.response?.status
-        ? t(
-            'genericcomponent.table.labels.notTwingraphDatasetError',
-            'Only twingraph datasets can be used to fetch table data dynamically'
+      setPlaceholderTitle(
+        error?.response?.statusText ?? t('genericcomponent.table.labels.queryFailedError', 'Query failed')
+      );
+      setPlaceholderBody(
+        error?.response?.data?.detail ??
```
esasova, line 271: maybe it's better to put the `error?.response?.data?.detail` in the console and use only custom error message in the placeholder? I'm afraid that sometimes, api can return huge error messages with technical details and it can break the layout

src/components/ScenarioParameters/components/ScenarioParametersInputs/GenericTable.js
```diff
@@ -209,76 +233,107 @@ export const GenericTable = ({
...
+    if (findDatasetById(sourceDatasetId)?.ingestionStatus == null) {
+      setPlaceholderTitle(t('genericcomponent.table.labels.noTwingraph', 'No Twingraph'));
+      setPlaceholderBody(
+        t(
+          'genericcomponent.table.labels.notTwingraphDatasetError',
+          'Only twingraph datasets can be used to fetch table data dynamically'
+        )
+      );
+      return;
+    }
```
esasova, line 318: I think it's better to put this error before the API call because the query will fail anyway so we don't need to send it if the dataset isn't twingraph

src/views/DatasetManager/components/DatasetOverview/components/CategoryAccordion/components/CategoryDetailsDialog/CategoryDetailsDialog.js
```diff
@@ -0,0 +1,162 @@
...
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [open, clearApplicationErrorMessage]); // Do not include applicationError in deps list
+
+  const handleOpen = useCallback(() => {
+    setOpen(true);
+  }, []);
+
+  const handleClose = useCallback(() => {
+    if (previousAppError.current != null) setApplicationErrorMessage(previousAppError.current);
+    else clearApplicationErrorMessage();
```
esasova, line 43: Just a question: now, when all errors are handled in the placeholder, do we still need this logic with previous errors?

src/views/DatasetManager/components/DatasetOverview/components/CategoryAccordion/components/CategoryDetailsDialog/CategoryDetailsDialog.js
```diff
@@ -0,0 +1,162 @@
...
+          fullScreen
+          open={open}
+          onClose={handleClose}
+          fullWidth={true}
+          maxWidth={'xl'}
+          PaperProps={{
+            elevation: 1,
+            style: {
+              minHeight: '100%',
+              maxHeight: '100%',
```
esasova, line 123: Do we need both properties ? `minHeight` alone isn't enough ? :thinking:

src/views/DatasetManager/components/DatasetOverview/components/CategoryAccordion/components/CategoryDetailsDialog/CategoryDetailsDialog.js
```diff
@@ -0,0 +1,162 @@
...
+            },
+          }}
+        >
+          <DialogTitle>
+            <ApplicationErrorBanner />
```
esasova, line 128: Do we still need this component here even with all errors handled in placeholder?

src/views/DatasetManager/components/DatasetOverview/components/CategoryAccordion/components/CategoryDetailsDialog/CategoryDetailsDialogHook.js
```diff
@@ -0,0 +1,38 @@
...
+import { useCurrentDataset, useCurrentDatasetId } from '../../../../../../../../state/hooks/DatasetHooks';
+import { useWorkspaceData } from '../../../../../../../../state/hooks/WorkspaceHooks';
+
+export const useCategoryDetailsDialogHook = () => {
+  const currentDataset = useCurrentDataset();
+  const currentDatasetId = useCurrentDatasetId();
+  const workspaceData = useWorkspaceData();
+  const isDarkTheme = useIsDarkTheme();
+  const applicationError = useApplicationError();
+  const setApplicationErrorMessage = useSetApplicationErrorMessage();
```
esasova, line 19: The same question about error banner :slightly_smiling_face:

cypress/e2e/brewery/DatasetManager_Overview.cy.js
```diff
@@ -0,0 +1,91 @@
+// Copyright (c) Cosmo Tech.
+// Licensed under the MIT license.
+import { Login, DatasetManager } from '../../commons/actions';
+import { TableParameters } from '../../commons/actions/generic';
```
esasova, line 4: Could you please shorten the import to` import { TableParameters } from '../../commons/actions` :pray: :slightly_smiling_face:

cypress/e2e/brewery/DatasetManager_Overview.cy.js
```diff
@@ -0,0 +1,91 @@
...
+
+const getTableTransport = () => cy.get(`[data-cy=table-transport]`);
+
+describe('Dataset manager overview work correctly', () => {
+  before(() => {
+    stub.start();
+    stub.setOrganizations([ORGANIZATION_WITH_DEFAULT_ROLE_USER]);
+    stub.setSolutions([SOLUTION]);
+    stub.setWorkspaces([WORKSPACE]);
+    stub.setDatasets([...DATASETS]);
```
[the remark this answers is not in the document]
nborde-CSM: Refacto, copy/paste mess...

src/components/ScenarioParameters/components/ScenarioParametersInputs/GenericTable.js
nborde-CSM: I thought it was used twice... Maybe a mess after refactor.

src/views/DatasetManager/components/DatasetOverview/components/CategoryAccordion/components/CategoryDetailsDialog/CategoryDetailsDialog.js
nborde-CSM: In fact, it seems that there is no need neither of the two props...

cypress/commons/actions/generic/DatasetManager.js
```diff
@@ -8,7 +8,11 @@ const SELECTORS = GENERIC_SELECTORS.datasetmanager;
 
 export const getDatasetManagerTab = (timeout = 5) => cy.get(SELECTORS.tabName, { timeout: timeout * 1000 });
 export const getDatasetManagerView = (timeout = 5) => cy.get(SELECTORS.view, { timeout: timeout * 1000 });
-export const switchToDatasetManagerView = () => getDatasetManagerTab().click();
+export const switchToDatasetManagerView = (queries = []) => {
+  const alias = api.interceptPostDatasetTwingraphQueries(queries, null, queries?.length ?? 1);
+  getDatasetManagerTab().click();
+  if (Array.isArray(queries) && queries?.length > 0) api.waitAliases(queries.map(() => alias));
```
csm-thu, line 14: The logic of handling the value of `queries` should probably be "hidden" in `interceptPostDatasetTwingraphQueries` instead of being implemented here  in `switchToDatasetManagerView`.

Could we make `interceptPostDatasetTwingraphQueries` return an array of aliases, and simply call `api.waitAliases` on this array? (if the array is empty, `api.waitAliases` will just ignore it and do nothing)

Also, I'm not sure how cypress will act when waiting multiple times for the same alias, but I think it's equivalent to waiting for it once. Do we really need to wait multiple aliases? (idem in `selectDatasetById` and `openCategoryDetailsDialog`)

cypress/commons/actions/generic/DatasetManager.js
```diff
@@ -90,13 +98,39 @@ export const getDatasetOverviewPlaceholderRetryButton = () => cy.get(SELECTORS.o
...
+export const getCategoryDetailsDialogOpenButton = (parent) =>
+  parent.find(SELECTORS.overview.categories.detailsDialogOpenButton);
+export const getDetailsDialogCategoryName = () => cy.get(SELECTORS.overview.categories.detailsDialogCategoryName);
+export const getDetailsDialogDatasetName = () => cy.get(SELECTORS.overview.categories.detailsDialogDatasetName);
+export const getCategoryDetailsDialogCloseButton = () => cy.get(SELECTORS.overview.categories.detailsDialogCloseButton);
 export const getKpiError = (parent) => parent.find(SELECTORS.overview.indicators.kpiError);
 export const getKpiUnknownState = (parent) => parent.find(SELECTORS.overview.indicators.kpiUnknownState);
 
+export const openCategoryDetailsDialog = (categoryId, query) => {
+  const alias = api.interceptPostDatasetTwingraphQuery(query.results);
```
csm-thu, line 127: If the function doesn't use the query object but only `query.results`, then maybe we should put directly `results` as function argument :slightly_smiling_face:

PR #902, opened 2024-08-27
feat: [PROD-12789] add dataset name in scenario view

esasova:
- [ ] Code is clean

### DIFF
--- a/src/views/Scenario/ScenarioHook.js
+++ b/src/views/Scenario/ScenarioHook.js
@@ -1,6 +1,9 @@
 // Copyright (c) Cosmo Tech.
 // Licensed under the MIT license.
+import { useMemo } from 'react';
+import { useTranslation } from 'react-i18next';
 import { useSetApplicationErrorMessage } from '../../state/hooks/ApplicationHooks';
+import { useDatasets } from '../../state/hooks/DatasetHooks';
 import { useOrganizationId } from '../../state/hooks/OrganizationHooks';
 import {
   useCurrentScenarioLastRunId,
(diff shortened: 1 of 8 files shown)


src/views/Scenario/ScenarioHook.js
```diff
@@ -17,7 +19,10 @@ export const useScenario = () => {
   const currentScenarioData = useCurrentScenarioData();
   const organizationId = useOrganizationId();
   const workspaceId = useWorkspaceId();
-
+  const datasets = useWorkspaceDatasets();
```
csm-thu, line 22: `useWorkspaceDatasets` may not return all datasets, I think it's possible to have scenarios that use datasets that won't be in this list (either because the workspace filter has changed, or because the scenario has been created manually with the API). Maybe we should retrieve the dataset list from `useDatasets` instead?

src/views/Scenario/ScenarioHook.js
```diff
@@ -17,7 +19,10 @@ export const useScenario = () => {
...
+  const currentScenarioDatasetName = useMemo(() => {
+    return datasets?.find((dataset) => dataset.id === currentScenarioData?.datasetList[0])?.name;
```
csm-thu, line 24: Not sure if we want to hide it when we don't find the dataset id in the list, but we could add a fallback value here (e.g. `N/A`, `Not found`, ...) to explain that we didn't find the dataset id? (e.g. if the dataset has been deleted).

Also, the part `datasetList[0]` may crash the webapp in the specific case of scenarios that have been created manually without any datasets. Maybe we should display a different message in this case (e.g. `No dataset`) or hide the dataset name label in this case?

src/views/Scenario/ScenarioHook.js
esasova: Good idea ! :+1:

csm-thu (approved):
LGTM :+1:

PR #931, opened 2024-10-08
Esas/migrate to react 18

esasova:
- [ ] Code is clean


nborde-CSM (approved):
LGTM :+1:

src/state/runner/reducers.js
```diff
@@ -0,0 +1,167 @@
+// Copyright (c) Cosmo Tech.
+// Licensed under the MIT license.
+// Current Simulation Runner
+import { createSlice } from '@reduxjs/toolkit';
+import { STATUSES } from '../../services/config/StatusConstants';
+import { RunnersUtils } from '../../utils';
+
+// Runners List
```
csm-thu, line 8: Not sure this comment is still relevant, because the runner list and the "current runner" are all in the same state

src/state/runner/sagas/PollRunnerState.js
```diff
@@ -97,7 +101,7 @@ function* startPolling(action) {
   stopActionName += '_' + action.runnerId;
 
   // Prevent double polling on same scenario
-  yield put(forgeStopPollingAction(action.runnerId));
+  // yield put(forgeStopPollingAction(action.runnerId));
```
csm-thu, line 104: If this line is no longer required, then we should probably remove it :smile:

csm-thu (approved):
Minor changes suggested, btu LGTM :+1:

src/state/runner/sagas/PollRunnerState.js
esasova: That's strange, I tried to delete this action and it broke polling, I will retry it one more time to see why it's not broken this time :sweat_smile:

esasova: Yes, we still need it to avoid a loop of update in Redux, I'll fix it, thanks ! :pray:

esasova: UPD: we still need this action on line 92 when the run stops but not on this line where it prevents the polling on scenario change but isn't necessary for runners as we don't fetch the runner by id on scenario change

PR #932, opened 2024-10-09
fix: [PROD-13974] fix not updated parentId when a runner is deleted

esasova:
- [ ] Code is clean

### DIFF
--- a/src/utils/RunnersUtils.js
+++ b/src/utils/RunnersUtils.js
@@ -32,7 +32,17 @@ const patchRunnerWithCurrentUserPermissions = (runner, userEmail, userId, permis
   };
 };
 
+const updateParentIdOnDelete = (runners, deletedRunnerId) => {
+  const parentId = runners.find((runner) => runner.id === deletedRunnerId)?.parentId;
+  runners.forEach((runner) => {
+    if (runner.parentId !== deletedRunnerId) return;
+
+    runner.parentId = parentId ?? null;
+  });
+};
+
 export const RunnersUtils = {
   patchRunnerWithCurrentUserPermissions,
   patchRunnerParameterValues,
+  updateParentIdOnDelete,
 };
--- a/src/utils/__test__/RunnersUtils.spec.js
+++ b/src/utils/__test__/RunnersUtils.spec.js
@@ -61,3 +61,100 @@ describe('patchScenarioParameterValues', () => {
     spyConsoleWarn.mockClear();
   });
 });
+
+describe('update parentId on delete', () => {
+  test('can update parentId if parent has been deleted', () => {
+    const runners = [
+      { id: 'r-1', parentId: null },
+      { id: 'r-2', parentId: 'r-1' },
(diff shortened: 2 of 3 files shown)


src/utils/RunnersUtils.js
```diff
@@ -32,7 +32,20 @@ const patchRunnerWithCurrentUserPermissions = (runner, userEmail, userId, permis
...
+const updateParentIdOnDelete = (runners, deletedRunnerId) => {
+  const parentId = runners.find((runner) => runner.id === deletedRunnerId)?.parentId;
+  const childrenIds = runners.filter((runner) => runner.parentId === deletedRunnerId)?.map((runner) => runner.id);
+  if (childrenIds.length > 0) {
+    runners.forEach((runner) => {
+      if (childrenIds.includes(runner.id)) {
+        runner.parentId = parentId ?? null;
+      }
+    });
+  }
```
csm-thu, line 44: Maybe I'm missing something, but I think we could simplify this by doing only one `forEach` loop without the first `filter` + `map`. Something like:
```
const parentId = runners.find((runner) => runner.id === deletedRunnerId)?.parentId;
runners.forEach((runner) => {
  if (runner.parentId !== deletedRunnerId) return; // Skip datasets that must remain unchanged
  runner.parentId = parentId ?? null;
});
```

src/utils/__test__/RunnersUtils.spec.js
```diff
@@ -61,3 +61,86 @@ describe('patchScenarioParameterValues', () => {
...
+    ];
+    const runnersAfterUpdate = [
+      { id: 'r-1', parentId: null },
+      { id: 'r-2', parentId: null },
+      { id: 'r-3', parentId: null },
+    ];
+    RunnersUtils.updateParentIdOnDelete(runners, 'r-1');
+    expect(runners).toStrictEqual(runnersAfterUpdate);
+  });
+  test("doesn't update array if no parent scenarios", () => {
```
csm-thu, line 94: Maybe the title should read "no children scenarios" instead of "no parent"?

src/utils/__test__/RunnersUtils.spec.js
```diff
@@ -61,3 +61,86 @@ describe('patchScenarioParameterValues', () => {
     spyConsoleWarn.mockClear();
   });
 });
+
+describe('update parentId on delete', () => {
```
csm-thu, line 65: Can we add a test to check the behavior when a scenario defined as `parentId` is not defined in the runners list? (this may happen when users have access to a child scenario but not its parent, because of ACL restrictions)

Example, with this structure of runners, but where `r-root` would not be shared with the current user:
```
 - `r-root`
   - `r-toremove`
     - `r-childscenario`
```

csm-thu (approved):

src/utils/RunnersUtils.js
esasova: Good idea, thanks ! :pray:

PR #948, opened 2025-01-28
fix: [PROD-14207]  fix deduplication of links when name attribute is not used

csm-thu:
In some twingraph instances, arcs and edges are identified by an 'id'
column, and not a 'name'. This fix makes the use of the id column the
new default behavior, because ids are supposed to be uniques (whereas
entity and link names can be reused).
If no id is defined for links, the webapp will try to fall back to
the previous behavior by looking for the 'name' attribute.


esasova (approved):
LGTM :+1:

PR #952, opened 2025-02-27
Thu/edit and relaunch dataset runners prod 14182

esasova:
- [ ] Code is clean


src/views/DatasetManager/components/CreateDatasetButton/components/DatasetCreationParameters/DatasetCreationParameters.js
```diff
@@ -107,7 +101,7 @@ export const DatasetCreationParameters = ({ dataSourceRunTemplates, parentDatase
                   gridItemProps={{ xs: 12, sx: { pt: 2 } }}
                   parameterData={parameter}
                   context={{ editMode: true, targetDatasetId: parentDataset?.id }}
-                  parameterValue={value}
+                  parameterValue={Array.isArray(value) ? value : value?.split(',')}
```
csm-thu, line 104: I wonder why we have to parse the string value here, whereas the other types of parameters don't need any conversion. We even have a [conversion mechanism for the list type](https://github.com/Cosmo-Tech/azure-sample-webapp/blob/experimental/on_premises_api_v4/src/utils/scenarioParameters/generic/ConversionFromString.js#L48), so it should always be stored as an Array in memory, right?

src/views/DatasetManager/components/CreateDatasetButton/components/DatasetCreationParameters/DatasetCreationParameters.js
```diff
@@ -183,36 +177,45 @@ export const DatasetCreationParameters = ({ dataSourceRunTemplates, parentDatase
         };
   }, [t, isSubDatasetCreationWizard]);
 
+  const sourceTypeComponent =
+    selectedRunner && Object.keys(selectedRunner).length > 0 ? (
```
csm-thu, line 181: Maybe we could create a bool constant with an easy-to-read name to replace both this expression `selectedRunner && Object.keys(selectedRunner).length > 0` and its opposite `!selectedRunner || Object.keys(selectedRunner)?.length === 0` that is used in the `return` statement

src/views/DatasetManager/components/UpdateDatasetButton/UpdateDatasetButton.js
```diff
@@ -0,0 +1,60 @@
...
+
+export const UpdateDatasetButton = ({ dataset }) => {
+  const { t } = useTranslation();
+  const runners = useGetAllRunners();
+  const selectedRunner = runners.find((runner) => runner.id === dataset.source.name);
+
+  const isDisabled =
+    (dataset && dataset.ingestionStatus === INGESTION_STATUS.PENDING) ||
+    !selectedRunner ||
+    Object.keys(selectedRunner ?? {})?.length === 0;
```
csm-thu, line 22: `!selectedRunner` seems to be redundant with `Object.keys(selectedRunner ?? {})?.length === 0`, I think we could keep the second one and remove the part `!selectedRunner` :slightly_smiling_face: 
Also, `dataset && dataset.ingestionStatus === INGESTION_STATUS.PENDING` could probably be simplified into `dataset?.ingestionStatus === INGESTION_STATUS.PENDING`

csm-thu (approved):
Suggested some changes, but LGTM :+1:

src/state/hooks/RunnerHooks.js
```diff
@@ -25,3 +25,18 @@ export const useStopRunner = () => {
     [dispatch, organizationId, workspaceId]
   );
 };
+
+export const useGetAllRunners = () => {
```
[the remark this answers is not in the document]
esasova: You are right, this hook returns only ETL runners so it must be called `useGetETLRunners` :+1: even in webapp v7, this will be the case because only initial saga will retrieve all runners, in components, we will have to separate simulation runners from ETL ones.

src/views/DatasetManager/components/CreateDatasetButton/components/DatasetCreationParameters/DatasetCreationParameters.js
esasova: I honestly don't remember why I did it :thinking:

csm-thu (approved):
LGTM :+1:

nborde-CSM (commented):
One question but LGTM :+1:

src/state/reducers/runner/RunnerReducer.js
```diff
@@ -0,0 +1,30 @@
+// Copyright (c) Cosmo Tech.
+// Licensed under the MIT license.
+import { createReducer } from '@reduxjs/toolkit';
+import { RUNNER_ACTIONS_KEY } from '../../commons/RunnerConstants';
+
+const initialState = {
+  etlRunners: [],
```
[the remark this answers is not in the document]
csm-thu: That's open to discussion, but I'd suggest we (almost) always use camelCase for acronyms, in full lowercase if they're at the beginning of the word :slightly_smiling_face: 
Examples:
- `etlRunners`
- `getEtlRunners()`

I think it makes it easier to read the words, even when there are several consecutive acronyms (e.g. `csmApiEtlId` vs. `CSMAPIETLID`).

It may also help case-conversion tools & functions. For example, a tool/function converting camelCase to kebab-case could easily translate `csmApiEtlId` to `csm-api-etl-id`, but it would be impossible to find the same from `CSMAPIETLID`.

PR #957, opened 2025-03-24
fix: fix broken theme switch button

csm-thu:
- app reducer 'setApplicationTheme' now expects an object payload with a field `isDarkTheme`, and not just the boolean value
- moved the call to `setApplicationTheme` in the `useEffect` triggered by the change of the state `isDarkThemeUsed`, so we can use the function-shaped setter: `setIsDarkThemeUsed((previousIsDarkThemeUsed) => !previousIsDarkThemeUsed)`


esasova (approved):
LGTM :+1:

PR #961, opened 2025-04-03
Esas/disable scenario run prod 13346

esasova:
- [ ] Code is clean

### DIFF
--- a/public/locales/en/translation.json
+++ b/public/locales/en/translation.json
@@ -41,6 +41,7 @@
           "discard": "DISCARD",
           "edit": "Edit parameters",
           "launch": "LAUNCH",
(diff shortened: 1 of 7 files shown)


public/locales/en/translation.json
```diff
@@ -41,6 +41,7 @@
           "discard": "DISCARD",
           "edit": "Edit parameters",
           "launch": "LAUNCH",
+          "launchDisabled": "The scenario cannot be run because its dataset isn't found or has ingestion status ERROR",
```
csm-thu, line 44: I think the "ERROR" status is slightly too technical to be displayed in the error message. Maybe we can rephrase the message into "[...] or its data ingestion has failed"? (same thing for the french translation, and the default string fallback in the component)

PR #963, opened 2025-04-24
fix: fix "stop run" button that was sometimes sending the id of the previous run

csm-thu:
- "Abort" button is now disabled when run id of the current scenario is nullish
- when starting a new run, the redux saga locally erases the scenario run data to force the scenario run to be nullish


esasova (approved):
LGTM :+1:

PR #964, opened 2025-04-24
fix: [PROD-14341] fix missing data in instance view inspector

csm-thu:
In some twingraph instances, for a given category of node, there can exist
several "groupes" of nodes, with different attributes. When this is the case,
the webapp query that retrieves the list of keys for each category will
return multiple items for the same category (one for each distinct set of
node attributes). The previous behavior of the webapp kept only the last one
of these items, thus possibly losing some keys in the process. The following
cypher query fetching the nodes information would then not ask for all the
existing properties, leading to missing data in the inspector, in the
instance view.

The fix creates a dictionary of node keys by category to regroup all possibly
existing properties before fetching the nodes data. Edges were impacted by
the same bug, so the same fix has been implemented.


esasova (approved):
LGTM :+1:

PR #971, opened 2025-05-21
refactor: upgrade to PIXJS v8

nborde-CSM:
- [ ] Code is clean


package.json
```diff
@@ -14,7 +14,6 @@
     "@mui/lab": "^5.0.0-alpha.175",
     "@mui/material": "^5.17.1",
     "@mui/styles": "^5.17.1",
-    "@pixi/unsafe-eval": "^7.4.3",
```
csm-thu, line 17: Removing this dependency broke the built version of the webapp (the graph is not visible in the Simulation view), with this error in the console:
> Uncaught (in promise) Error: Current environment does not allow unsafe-eval, please use pixi.js/unsafe-eval module to enable support.

src/views/Simulation/components/SceneContainer/SceneContainer.js
```diff
@@ -36,9 +27,7 @@ const SceneContainer = ({ toggleInspectorDrawer }) => {
         data-cy="pixi-d3-view"
         ref={containerRef}
         style={{ width: '100%', height: '100%', display: 'flex', flexDirection: 'column' }}
-      >
-        <div ref={canvasRef} style={{ flex: 1, width: '100%', position: 'relative', overflow: 'hidden' }}></div>
-      </div>
```
csm-thu, line 41: Do we still need the parent div if the div with `canvasRef` is removed?

src/views/Simulation/components/SceneContainer/pixiUtils.js
```diff
@@ -1,57 +1,65 @@
...
 
-const createTexture = (appRef, options) => {
-  const graphics = new PIXI.Graphics();
-  graphics.beginFill(options.fillColor);
-  graphics.lineStyle(0.75, options.lineColor);
-  graphics.drawRect(0, 0, 8, 8);
-  graphics.endFill();
-  return appRef.current.renderer.generateTexture(graphics, PIXI.SCALE_MODES.LINEAR, 2);
+const createGraphicsContext = (options) => {
+  const grph = new GraphicsContext()
```
csm-thu, line 7: Do we really need the intermediate `grph` const in this function?
If we need it, then we should probably rename it and use full words for variable names, for readability purposes :slightly_smiling_face:

src/views/Simulation/components/SceneContainer/SceneContainer.js
```diff
@@ -3,30 +3,21 @@
...
-import { createApp, renderElements } from './pixiUtils';
+import { createApp, destroyApp, initApp } from './pixiUtils';
 
 const SceneContainer = ({ toggleInspectorDrawer }) => {
   const theme = useTheme();
 
   const containerRef = useRef(null);
-  const canvasRef = useRef(null);
-  const appRef = useRef(null);
+  const app = createApp();
```
csm-thu, line 13: Calling `createApp` in the main render function of the component will create a new Pixi `Application` every time the component is re-rendered. Also, these new applications will never be used because `initApp` is called only once when the component is mounted for the first time (in the `useEffect` with the empty dependency array)

package.json
nborde-CSM: pixi/unsafe-eval is now directly integrated with PixiJS v8

src/views/Simulation/components/SceneContainer/pixiUtils.js
```diff
@@ -1,57 +1,65 @@
...
 import { getGraphFromInstance } from './graphUtils';
 
-const createTexture = (appRef, options) => {
-  const graphics = new PIXI.Graphics();
-  graphics.beginFill(options.fillColor);
-  graphics.lineStyle(0.75, options.lineColor);
-  graphics.drawRect(0, 0, 8, 8);
-  graphics.endFill();
-  return appRef.current.renderer.generateTexture(graphics, PIXI.SCALE_MODES.LINEAR, 2);
+const createGraphicsContext = (options) => {
```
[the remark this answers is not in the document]
nborde-CSM: For the future representation of nodes, I think we will not directly use icon + graphics but use combos of sprites.

src/views/Simulation/components/SceneContainer/SceneContainer.js
nborde-CSM: Initially, I got out the app creation outside the `useEffect` because I thought it caused screen app duplication during each app live update. Later I discovered it was the `useRef` use issue. I forgot to replace the app creation inside the `useEffect`. It's done now.

PR #973, opened 2025-05-23
feat: [PROD-14412] add random values in KPI cards

csm-thu:


src/views/Simulation/components/Kpis/KpiCard.js
```diff
@@ -0,0 +1,37 @@
...
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Card, Grid, Typography } from '@mui/material';
+import KpiTrendChip from './KpiTrendChip';
+
+const KpiCard = ({ kpi }) => {
+  return (
+    <Card sx={{ width: '100%', p: 2 }}>
+      <Grid container justifyContent="space-between" alignItems="center" columnSpacing={2}>
+        <Grid item xs="auto">
```
nborde-CSM, line 12: `xs="auto"` is the same thing as `xs` ?

src/views/Simulation/components/Kpis/KpiCard.js
```diff
@@ -0,0 +1,37 @@
...
+            {kpi?.label}
+          </Typography>
+        </Grid>
+      </Grid>
+    </Card>
+  );
+};
+
+KpiCard.propTypes = {
+  kpi: PropTypes.object,
```
nborde-CSM, line 34: Maybe this prop should be _required_ ?

nborde-CSM (approved):
Some remarks but LGTM :+1:

src/views/Simulation/components/Kpis/KpiCard.js
csm-thu: Not exactly, according to [the documentation](https://v5.mui.com/material-ui/react-grid/#variable-width-content):
> Set one of the size breakpoint props to "auto" instead of true / a number to size a column based on the natural width of its content.

PR #976, opened 2025-06-04
feat: [PROD-14410] implement top bar & settings pop-up in Simulation view

csm-thu:


src/views/Simulation/utils/pixiUtils.js
```diff
@@ -77,17 +84,19 @@ export const createApp = async (containerRef, sceneContainerRef, instance, theme
   app.stage.addChild(sceneContainerRef.current);
 
   const handleResize = () => {
-    if (!containerRef.current || !app) return;
+    if (!containerRef.current || !app?.renderer) return;
```
nborde-CSM, line 87: Is it really usefull to use `app?.render` instead of `app` ?

nborde-CSM (approved):
One question but LGTM :+1:

src/views/Simulation/utils/pixiUtils.js
csm-thu: In some cases, the app seems to exist whereas the renderer inside no longer exists. This might be because of the hot reload when using a local webapp, I don't know, but checking for `!app?.renderer` will cover both cases (when the app does not exist, and when it exists but the renderer no longer exists)

PR #989, opened 2025-06-24
Thu/improve perf for filter settings

csm-thu:


src/views/Simulation/utils/graphUtils.js
```diff
@@ -99,7 +99,27 @@ const setResourceBottlenecks = (instance, productionResources, bottlenecks, sett
...
+  const defaultGrayedOutValue = !showAllElements;
+
+  graph.links.forEach((link) => (link.isGrayedOut = defaultGrayedOutValue));
+  graph.nodes.forEach((node) => (node.isGrayedOut = defaultGrayedOutValue));
+  if (showAllElements) return;
+
+  const highlightBottlenecks = settings.graphViewFilters.includes(GRAPH_VIEW_FILTER_VALUES.BOTTLENECKS);
+  const highlightShortages = settings.graphViewFilters.includes(GRAPH_VIEW_FILTER_VALUES.SHORTAGES);
+  graph.nodes.forEach((node) => {
+    if (highlightShortages && (node.shortagesCount ?? 0) > 0) node.isGrayedOut = false;
```
nborde-CSM, line 113: Suggestion :
For more readability, we can write :
`const shortagesCount = node.shortagesCount ?? 0;
if (highlightShortages && shortagesCount > 0)`

src/views/Simulation/utils/graphUtils.js
```diff
@@ -99,7 +99,27 @@ const setResourceBottlenecks = (instance, productionResources, bottlenecks, sett
...
+    if (highlightBottlenecks && (node.bottlenecksCount ?? 0) > 0) node.isGrayedOut = false;
```
nborde-CSM, line 114: Same suggestion

nborde-CSM (approved):
One suggestion but LGTM :+1:

src/views/Simulation/utils/graphUtils.js
csm-thu: You are right, I agree that this would be more readable, and I would do this change for most other parts of the code. 

Yet, for the specific case of a PR targeting the application performance, in a loop on all graph nodes, in a part of code that have been identified as having a measurable impact on performance, I would say that this case is an exception where we can afford losing some readability. Furthermore, the code is still simple and remains readable enough, IMO.

We could do even better (for the performance) by omitting the fallback value and writing:
`if (highlightShortages && node.shortagesCount > 0) node.isGrayedOut = false;`
because `undefined` compared to `0` gives `false`, but it starts feeling like it could hide some bugs, so we should probably keep the current version instead :smile: 

Let's meet and discuss this if you want :slightly_smiling_face:

PR #995, opened 2025-07-07
fix: [PROD-14488] fix focus and zoom when selecting element on scene

nborde-CSM:
- [ ] Code is clean

### DIFF
--- a/src/views/Simulation/utils/SceneContainer.js
+++ b/src/views/Simulation/utils/SceneContainer.js
@@ -6,6 +6,7 @@ const MIN_ZOOM = 0.05;
 const MAX_ZOOM = 2;
 const DEFAULT_ZOOM = 0.2;
 const ZOOM_SPEED = 0.5;
+const FOCUS_ZOOM = 1;
 
 const interpolateLinear = (a, b, t) => {
   return a + (b - a) * t;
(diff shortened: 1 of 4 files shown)


src/views/Simulation/utils/SceneContainer.js
```diff
@@ -96,6 +97,49 @@ export class SceneContainer extends Container {
...
+  findElementById(elementId) {
+    const recurseSearch = (elements) => {
+      if (!Array.isArray(elements) || elements.length === 0) return null;
+      let elementFound = elements.find((child) => child.elementId === elementId);
+      if (elementFound != null) return elementFound;
+
+      for (const child of elements) {
+        elementFound = recurseSearch(child.children);
+        if (elementFound != null) return elementFound;
+        else continue;
```
csm-thu, line 109: I think `else continue` is unnecessary here :thinking:

src/views/Simulation/utils/SceneContainer.js
```diff
@@ -96,6 +97,49 @@ export class SceneContainer extends Container {
...
+  getScreenCenterPoint() {
+    return new Point(this.canvasScene.clientWidth / 2, this.canvasScene.clientHeight / 2);
+  }
+
+  getPointWithZoomOffset(x, y, zoom = this.zoom) {
+    return new Point(x * zoom, y * zoom);
+  }
+
+  centerOnElement(elementId) {
+    const elementFound = this.findElementById(elementId);
```
csm-thu, line 125: Not a huge fan of having to search for an item in the containers to find its coordinates, it would be better to retrieve the element coordinates (computed by the layout) library, and simply translate them to the scene coordinates system. 
But I guess we can keep it like this for now. Performance shouldn't be a problem because this event is only triggered manually.

csm-thu (approved):
LGTM :+1:

src/views/Simulation/utils/SceneContainer.js
nborde-CSM: The element coordinates computed by the layout offer no guaranty to got the final coordinates elements after transformations and scales. Moreover, the graphics links need some computes to get an element center and I think it is quite hard to get it only from the layout coordinates.

PR #1017, opened 2025-08-06
fix: [PROD-14634] fix scene re-center issue and highlighning reset wh…

nborde-CSM:
…en changing scenario

- [ ] Code is clean

### DIFF
--- a/src/views/Simulation/components/Scene/Scene.js
+++ b/src/views/Simulation/components/Scene/Scene.js
@@ -80,9 +80,11 @@ const Scene = () => {
 
   useEffect(() => {
     if (!needsReRendering) return;
-    if (requiredUpdateStepsRef.current.all || requiredUpdateStepsRef.current.layout) {
+    const layoutUpdate = requiredUpdateStepsRef.current.all || requiredUpdateStepsRef.current.layout;
+    if (layoutUpdate) {
       resetGraphLayout(sceneCanvasRef.current.clientWidth, sceneCanvasRef.current.clientHeight);
     }
+    if (requiredUpdateStepsRef.current.all) setSelectedElementId(null);
 
     if (
       requiredUpdateStepsRef.current.all ||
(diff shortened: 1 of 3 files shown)


src/views/Simulation/components/Scene/Scene.js
```diff
@@ -80,8 +80,10 @@ const Scene = () => {
 
   useEffect(() => {
     if (!needsReRendering) return;
-    if (requiredUpdateStepsRef.current.all || requiredUpdateStepsRef.current.layout) {
+    const layoutUpdate = requiredUpdateStepsRef.current.all || requiredUpdateStepsRef.current.layout;
+    if (layoutUpdate) {
       resetGraphLayout(sceneCanvasRef.current.clientWidth, sceneCanvasRef.current.clientHeight);
+      setSelectedElementId(null);
```
csm-thu, line 86: This change means that the selected element will be reset every time the layout is changed? Can't we keep the selected element when we switch between horizontal and vertical layouts?

src/views/Simulation/components/Scene/Scene.js
```diff
@@ -92,7 +94,8 @@ const Scene = () => {
     ) {
       if (sceneContainerRef.current) sceneContainerRef.current.removeChildren();
       renderElements(sceneContainerRef, graphRef, setSelectedElementId, settings);
-      if (minimapContainerRef.current != null) minimapContainerRef.current?.renderElements();
+      if (layoutUpdate && sceneContainerRef.current) sceneContainerRef.current.setOrigin();
+      if (minimapContainerRef.current) minimapContainerRef.current?.renderElements();
```
csm-thu, line 98: > `if (minimapContainerRef.current) minimapContainerRef.current?.renderElements();`

Since you changed this line, is the optional chaining operator `?.` necessary here? The condition already covers the check, so we can change this part too

nborde-CSM: Yes, it is a remain after several changes, adds and re-changes...

arianber7 (approved):

PR #1027, opened 2025-08-20
Abe/dynamic view analyze & improve performance prod Chartimeline 14460

arianber7:
- [ ] Code is clean

### DIFF
--- a/src/views/Simulation/components/Charts/TimelineControls.js
+++ b/src/views/Simulation/components/Charts/TimelineControls.js
@@ -1,4 +1,4 @@
-import React, { useEffect, useRef, useState } from 'react';
+import React, { useEffect, useRef, useState, useCallback } from 'react';
 import PropTypes from 'prop-types';
 import CloseIcon from '@mui/icons-material/Close';
 import PauseRoundedIcon from '@mui/icons-material/PauseRounded';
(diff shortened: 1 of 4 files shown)


src/views/Simulation/components/Charts/TimelineControls.js
```diff
@@ -107,59 +133,40 @@ const TimelineControls = ({ chartData, markers, startDate, endDate }) => {
...
     }
-  };
+  }, [playbackSpeed]);
 
-  const handleChartClick = (e) => {
-    if (e?.activeLabel != null) {
-      const newStep = e.activeLabel;
-      setCurrentTimestep(newStep);
+  const handleChartClick = useCallback(
+    (e) => {
```
csm-thu, line 152: Using variable names as short as `e` will probably decrease readability for the other developers :)
We can usually find short and understandable names (e.g. `event`), and in other cases, we'd rather have a long variable (or function) name than one that is too short and could cause confusion :smile:

src/views/Simulation/components/Charts/TimelineControls.js
```diff
@@ -94,9 +96,33 @@ const TimelineControls = ({ chartData, markers, startDate, endDate }) => {
   const animationFrameId = useRef(null);
   const startTimeRef = useRef(0);
   const pausedAtRef = useRef(0);
-  const isPlayingRef = useRef(false); // live flag
+  const isPlayingRef = useRef(false);
+
+  const stepDurationRef = useRef(500);
```
csm-thu, line 101: `stepDurationRef` value is never changed, maybe we could move it in a `const` above the component?
(e.g. define `const STEP_DURATION_IN_MS = 500;` just after `SPEEDS` definition)

PR #1031, opened 2025-08-27
feat: [PROD-14648] implement dashed line effect on the required links

arianber7:
- [ ] Code is clean


src/views/Simulation/utils/pixiUtils.js
```diff
@@ -203,6 +203,65 @@ const createProductionResourceContainer = (textures, name, isHighlighted, operat
...
+  arrowGraphics.poly([0, 0, -size, size / 2, -size, -size / 2]);
+  arrowGraphics.endFill();
+  arrowGraphics.x = midX;
+  arrowGraphics.y = midY;
+  arrowGraphics.rotation = angle;
+
+  stage.addChild(arrowGraphics);
+}
+
+function drawDashedBezier(ctx, x1, y1, cp1x, cp1y, cp2x, cp2y, x2, y2, dash = 8, gap = 5) {
```
csm-thu, line 230: `drawDashedBezier` is only used once, with the values `dash = 4` and `gap = 4`. Maybe we could use these default values instead of 8 and 5? :slightly_smiling_face:

src/views/Simulation/utils/pixiUtils.js
```diff
@@ -236,22 +296,58 @@ const createLinkGraphics = (links, setSelectedElementId, settings) => {
...
-    graphics.bezierCurveTo(
-      controlPoint1.x,
-      controlPoint1.y,
-      controlPoint2.x,
-      controlPoint2.y,
-      target.x - targetOffset.x,
-      target.y - targetOffset.y
-    );
+
+    if (link.target.type === 'stock') {
```
csm-thu, line 300: Based on the mockups, I think we could use instead:
`if (link.type !== 'transports') {`

We already have the info in `link.type`, nothing to change in the graph processing functions. The existing link types are `input`, `output` and `transports`.

csm-thu (approved):

Good work!

PR #1035, opened 2025-09-03
fix: [PROD-14667] scene highliting not updated

arianber7:
- [ ] Code is clean


src/views/Simulation/utils/graphUtils.js
```diff
@@ -4,16 +4,17 @@ import { GRAPH_VIEW_FILTER_VALUES } from '../constants/settings';
...
 
 const forgeElementData = (element, keysToHide = []) => {
-  const data = { ...element };
-  keysToHide.forEach((keyToHide) => {
-    if (keyToHide in data) delete data[keyToHide];
-  });
+  const data = {};
+  for (const key in element) {
+    if (!keysToHide.includes(key)) data[key] = element[key];
+  }
```
csm-thu, line 10: I'm not sure I understand the reason for this change. If the goal was performance improvements, then I'm not sure this new version would be better, because the number of keys to hide should be much less than the number of keys to copy, so it seemed than copying the whole object and then deleting 2 or 3 keys should be faster, but maybe I missed something :thinking:

PR #1041, opened 2025-09-15
perf: [PROD-14741] add toogle button in flowchart settings to disable…

nborde-CSM:
… filters effects

- [ ] Code is clean

### DIFF
--- a/src/views/Simulation/SimulationViewHook.js
+++ b/src/views/Simulation/SimulationViewHook.js
@@ -93,6 +93,7 @@ export const useSimulationView = () => {
     settings.outputLevels,
     settings.orientation,
     settings.spacing,
+    settings.enableGlowEffect,
     selectedElementId,
   ]);
 
--- a/src/views/Simulation/components/TopBar/SettingsButton.js
+++ b/src/views/Simulation/components/TopBar/SettingsButton.js
@@ -5,7 +5,6 @@ import { SettingsOutlined as SettingsOutlinedIcon } from '@mui/icons-material';
 import {
   Box,
   Button,
-  Divider,
   FormControl,
   FormControlLabel,
   Menu,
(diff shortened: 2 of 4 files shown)


src/views/Simulation/SimulationViewHook.js
```diff
@@ -93,6 +93,7 @@ export const useSimulationView = () => {
     settings.outputLevels,
     settings.orientation,
     settings.spacing,
+    settings.filterEffectActivated,
```
csm-thu, line 96: The property name can be confusing, because of the other setting named `graphViewFilters`. Maybe we can rename `filterEffectActivated` to `enableGlowEffect`?

src/views/Simulation/components/TopBar/SettingsButton.js
```diff
@@ -191,6 +191,16 @@ export const SettingsButton = () => {
         {outputLevelsSelect}
       </Stack>
     );
+    const filterEffectsSettings = (
```
csm-thu, line 194: The "Settings" suffix is used above for a `Stack` of several components, but this one only has one `Switch` toggle. 
Can we rename it to `glowEffectSwitch`?

src/views/Simulation/utils/pixiUtils.js
```diff
@@ -134,13 +134,13 @@ const createFactoryIconGraphicsContext = (options) => {
...
 
-const createStockContainer = (textures, name, isHighlighted = false) => {
+const createStockContainer = (textures, name, isHighlighted = false, settings) => {
   const container = new Container();
   container.label = isHighlighted ? NODE_TYPES.STOCK_SHORTAGE : NODE_TYPES.STOCK;
   const stockTextureKey = isHighlighted ? 'stockLevel1' : 'stockLevel0';
   if (isHighlighted) {
     const stockHalo = new Sprite(textures[stockTextureKey]);
-    stockHalo.filters = [BLOOM_FILTER];
+    stockHalo.filters = settings.filterEffectActivated ? [BLOOM_FILTER] : [];
```
csm-thu, line 143: Instead of providing the whole `settings` object, could we replace it by a simple boolean value `enableGlowEffect` to keep this function low-level and focused on the Pixi-related logic?
Same thing for the functions `createProductionResourceContainer` and `createNodeContainer` if it's possible

PR #1054, opened 2025-10-03
feat: [PROD-14705] map view filters

arianber7:
- [ ] Code is clean

### DIFF
--- a/src/views/Simulation/SimulationViewHook.js
+++ b/src/views/Simulation/SimulationViewHook.js
@@ -40,6 +40,7 @@ export const useSimulationView = () => {
   const [centerToPosition, setCenterToPosition] = useState(() => {});
 
   const [viewMode, setViewMode] = useState(DEFAULT_SETTINGS.viewMode);
+  const [selectedMapFilters, setSelectedMapFilters] = useState([]);
 
   const graphRef = useRef(null);
   const lastScenarioId = useRef(null);
@@ -127,5 +128,7 @@ export const useSimulationView = () => {
     requiredUpdateStepsRef,
     viewMode,
     setViewMode,
+    selectedMapFilters,
+    setSelectedMapFilters,
   };
 };
--- a/src/views/Simulation/components/TopBar/MapEntities.js
+++ b/src/views/Simulation/components/TopBar/MapEntities.js
@@ -1,0 +1,153 @@
+import React from 'react';
+import {
+  Checkbox,
+  MenuItem,
+  FormControl,
+  Select,
+  ListItemText,
+  InputLabel,
+  Chip,
+  Box,
+  Typography,
+  Switch,
+} from '@mui/material';
+import { useSimulationViewContext } from '../../SimulationViewContext';
+import { MAP_VIEW_FILTERS } from '../../constants/settings';
+
+export const MapEntitiesDropdown = () => {
+  const { selectedMapFilters, setSelectedMapFilters } = useSimulationViewContext();
+  const onlyCriticalPointsChecked = selectedMapFilters.includes('onlyCriticalPoints');
+
+  const handleChange = (event) => {
+    const {
+      target: { value },
+    } = event;
+
+    const switchValues = selectedMapFilters.filter((v) =>
+      MAP_VIEW_FILTERS.some((f) => f.type === 'switch' && f.value === v)
+    );
+
+    const checkboxValues = MAP_VIEW_FILTERS.filter((f) => f.type === 'checkbox').map((f) => f.value);
+
+    const lastClicked =
+      value.find((v) => !selectedMapFilters.includes(v)) || selectedMapFilters.find((v) => !value.includes(v));
+
+    const addAllIfNeeded = (arr) => {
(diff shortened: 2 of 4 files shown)


src/views/Simulation/components/TopBar/MapEntities.js
```diff
@@ -0,0 +1,97 @@
...
+        />
+        <Chip label={`+${moreCount}`} size="small" sx={{ ml: 1 }} color="warning" />
+      </Box>
+    );
+  };
+
+  return (
+    <Box sx={{ display: 'flex', alignItems: 'center', gap: 2 }}>
+      <FormControl fullWidth variant="outlined" size="small">
+        <InputLabel id="map-entities-label">Entities</InputLabel>
```
csm-thu, line 66: Not sure why, but the "_Entities_" placeholder is not vertically centered in the input field:

Image

src/views/Simulation/constants/settings.js
```diff
@@ -15,6 +15,14 @@ export const SIMULATION_MODES = {
   MAP: 'map',
 };
 
+export const MAP_VIEW_FILTERS = [
+  { value: 'all', label: 'All' },
+  { value: 'production_resources', label: 'Production ressources' },
```
csm-thu, line 20: I think this might be a typo in the mockups: the label "Production ressources" should be "Production resources" (with one 's')

src/views/Simulation/SimulationViewHook.js
```diff
@@ -40,6 +40,8 @@ export const useSimulationView = () => {
   const [centerToPosition, setCenterToPosition] = useState(() => {});
 
   const [viewMode, setViewMode] = useState(DEFAULT_SETTINGS.viewMode);
+  const [selectedMapFilters, setSelectedMapFilters] = useState([]);
+  const [onlyCriticalPoints, setOnlyCriticalPoints] = useState(false);
```
csm-thu, line 44: Do you think we could store the `onlyCriticalPoints` value in the `selectedMapFilters`? (it is a filter of the map view, so I suppose it would make sense 😄)
We could then extract it from the array in the `MapEntitiesDropdown` component, to keep a separate logic for the checkboxes vs. the toggle component for this option

src/views/Simulation/components/TopBar/MapEntities.js
```diff
@@ -0,0 +1,97 @@
...
+
+  const handleChange = (event) => {
+    const value = event.target.value;
+
+    if (value.includes('all')) {
+      setSelectedMapFilters(MAP_VIEW_FILTERS.map((e) => e.value));
+    } else {
+      setSelectedMapFilters(value);
+    }
+  };
```
csm-thu, line 29: I think there might be a bug in the dropdown list (but I'm not sure if this would be in this function ^^). Once the "all" option has been checked, it is impossible to "uncheck" the other boxes in the dropdown list (clicking on the Chip to remove it works though)

src/views/Simulation/components/TopBar/MapEntities.js
```diff
@@ -0,0 +1,97 @@
...
+    } else {
+      setSelectedMapFilters((prev) => prev.filter((v) => v !== value && v !== 'all'));
+    }
+  };
+
+  const renderValue = (selected) => {
+    if (selected.length === 0) return 'Entities';
+    if (selected.length === 1) {
+      const label = MAP_VIEW_FILTERS.find((opt) => opt.value === selected[0])?.label;
+      return label || 'Entities';
```
csm-thu, line 43: When only one option is selected, could we return it in a Chip component too? Otherwise, the component looks like a "single option" dropdown list, it might be confusing for users 🙂

csm-thu (changes requested):

src/views/Simulation/components/TopBar/MapEntities.js
csm-thu: @arianber7 there's still a bug in the dropdown list: when the "all" option is selected, trying to uncheck one of the other boxes lead to all boxes being unchecked. The expected behavior would be to have only this box unchecked, while all other boxes should remain checked.
Also, when the "all" option is selected, it disables the "only critical points option", but I think these two options should be independent: the "all" option should only impact the checkboxes

csm-thu (changes requested):

csm-thu (changes requested):
@arianber7 I think there are still minor bugs remaining with the chips now :sweat_smile: 

Here are the repro steps for the bugs I found, starting from an empty list:

1. open the dropdown, check the "All" checkbox, enable "only critical points", close the dropdown, click on the delete icon of the "All" chip -> the option "only critical points" is disabled whereas it should remain enabled

2. open the dropdown, enable "only critical points", check the "All" checkbox, close the dropdown, click on the delete icon of the "only critical points" chip -> the option "All" is no longer checked, whereas it should remain enabled

3. open the dropdown, check the "Transports" checkbox (a chip "Transports" is displayed above the dropdown), enable "only critical points" (the chip "Transports" is still displayed), check the "Stocks" checkbox -> the "Transports" chip is replaced by a "Only critical points" chip, even though the option was already selected (keeping the previous chip would be more consistent)

csm-thu (changes requested):

PR #1134, opened 2026-06-12
fix: [PROD-16010] added options to query on table parameter

selmincosmotech:
- [ ] Code is clean

### DIFF
--- a/src/components/ScenarioParameters/components/ScenarioParametersInputs/GenericTable.jsx
+++ b/src/components/ScenarioParameters/components/ScenarioParametersInputs/GenericTable.jsx
@@ -272,7 +272,9 @@ export const GenericTable = ({
       return;
     }
 
-    const { datasetPartName, queryOptions } = ConfigUtils.getParameterAttribute(parameterData, 'dynamicValues');
+    const dynamicValuesAttr = ConfigUtils.getParameterAttribute(parameterData, 'dynamicValues') || {};
+    const datasetPartName = dynamicValuesAttr.datasetPartName;
+    const queryOptions = dynamicValuesAttr.options ?? {};
     const datasetPart = (sourceDataset.parts ?? []).find((part) => part.name === datasetPartName);
     if (datasetPart == null) {
       setPlaceholder({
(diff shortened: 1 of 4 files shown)


src/services/dataset/DatasetService.js
```diff
@@ -102,7 +102,8 @@ const queryDatasetPart = async (datasetPart, options = {}) => {
     offset,
     limit,
     groupBys,
-    orderBys
+    orderBys,
+    options
```
csm-thu: I don't think this change is necessary, the `options` object is already destructured to fill the parameters of the `queryData` function
According to [the docs](https://github.com/Cosmo-Tech/cosmotech-api-typescript-client/blob/master/docs/DatasetApi.md#querydata), `queryData` has no `options` argument

src/components/ScenarioParameters/components/ScenarioParametersInputs/GenericTable.jsx
```diff
@@ -272,7 +272,9 @@ export const GenericTable = ({
       return;
     }
 
-    const { datasetPartName, queryOptions } = ConfigUtils.getParameterAttribute(parameterData, 'dynamicValues');
+    const dynamicValuesAttr = ConfigUtils.getParameterAttribute(parameterData, 'dynamicValues') || {};
+    const datasetPartName = dynamicValuesAttr.datasetPartName;
+    const queryOptions = dynamicValuesAttr.queryOptions ?? dynamicValuesAttr.options ?? {};
```
csm-thu: I see, I think it was a mistake to have a property named `queryOptions` instead of `options`, maybe we should keep the simple destructuring
```
const { datasetPartName, options } = ConfigUtils.getParameterAttribute(parameterData, 'dynamicValues');
```
instead of trying to support both `queryOptions` and `options`.

And in [this file](https://github.com/Cosmo-Tech/azure-sample-webapp/blob/main/src/views/DatasetManager/components/DatasetOverview/components/CategoryAccordion/components/CategoryDetailsDialog/CategoryDetailsDialog.jsx#L54) we'll need to replace `queryOptions` by `options` in order to use the correct property name

csm-thu (approved):
LGTM 👍

